PR #6, opened 2017-07-20
Support for .NET Standard 2.0

vlesierse:
This pull request contains the changes need to make Qowaiv compatible with .NET Standard 2.0. This means it can run on .NET Core 2.0 and Xamarin.

- Moved Qowaiv, Qowaiv.Json.NewtonSoft and Qowaiv.Web to .NET Standard 2.0
- Version has been updated to 3.1.0
- AppVeyor is used as CI

*Note: Microsoft has released the Visual Studio 2017 update (15.3) which is needed to work with .NET Standard 2.0*

/cc: @Corniel


erik-am:
@Corniel What about merging this?

PR #7, opened 2017-10-02
Update info about Visual Studio requirements

Corniel:


vlesierse:
Apart from the code, which looks good to me, my suggestion is to include a download link to Visual Studio 2017 and the 15.3 Update.

PR #9, opened 2017-10-03
Introduced Qowaiv.ComponentModel

Corniel:
Thanks to @erik-am 

## 1. Annotations
We're extending the DataAnnotations from Microsoft with some more attributes of our own:

[Mandatory] Here the difference with Microsoft's [Required] attribute is that it works for Qowaiv Single Value Objects (and other value types following the same pattern) as well, it will be invalid if the default value is used
[AllowedValues] and [ForbiddenValues] make it easy to validate string values, or objects/value types that have a string representation

## 2. Result model
Also we propose a Result model that includes the validation messages, and Result<T> which can contain both an object and validation messages. This can be a helpful return type for methods that need to return objects but first have to validate them.

## 3. ValidationMessages with severity
The difference with Microsoft's default ValidationResult and our ValidationMessages is that in this PR we support a severity: info, warning, or error.


src/Qowaiv.ComponentModel/DataAnnotations/OptionalAttribute.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace Qowaiv.ComponentModel.DataAnnotations
+{
+	/// <summary>Decorates the item as being optional.</summary>
+	[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Parameter, AllowMultiple = false)]
+	public sealed class OptionalAttribute : RequiredAttribute
```
erik-am: This doesn't make sense, now an OptionalAttribute is also a RequiredAttribute.

src/Qowaiv.ComponentModel/Rules/Globalization/PostalCodeRule.cs
```diff
@@ -0,0 +1,56 @@
...
+            return PostalCode.IsEmptyOrUnknown() || Country.IsEmptyOrUnknown() || PostalCode.IsValid(Country);
+        }
+
+        /// <summary>Gets the error message.</summary>
+        protected override Func<string> ErrorMessageString => ()=> string.Format(QowaivComponentModelMessages.PostalCodeValidator_ErrorMessage, PostalCode, Country.DisplayName);
+
+		/// <summary>Returns <see cref="ValidationMessage.None"/> if the <see cref="PostalCode"/> 
+		/// is valid for a specific <see cref="Country"/>, otherwise false.
+		/// </summary>
+		public static ValidationResult IsValid(PostalCode postalCode, Country country, string postalCodeName, string countryName)
```
erik-am: tabs/spaces

src/Qowaiv.ComponentModel/Rules/ValidationRule.cs
```diff
@@ -0,0 +1,136 @@
...
+        protected ValidationRule(IEnumerable<string> propertyNames)
+        {
+            PropertyNames = (propertyNames ?? Enumerable.Empty<string>()).ToArray();
+        }
+
+        /// <summary>Gets the involved property names.</summary>
+        public string[] PropertyNames { get; }
+
+		/// <summary>Gets and set an error message to if validation fails.</summary>
+		public string ErrorMessage { get; set; }
```
erik-am: tabs/spaces

Issue #12, opened 2017-12-09
Base32 implementation under Creative Commons Attribution License


Corniel:
The base-32 implemenation (from: http://scottless.com/blog/archive/2014/02/15/base32-encoder-and-decoder-in-c.aspx) is licensed under the Creative Commons Attribution License. If I understand that license correctly, it is not possible to use this code under a MIT license. To prevent legal issues, the implementation should not longer rely on this (referred) piece of code.

PR #13, opened 2017-12-09
New implementation of base-32.

Corniel:
Key difference in implementation is that for both the ToString loop as the GetBytes loop are based on the 5bit char part instead of the 8bit byte part.

This should also solve the legal issue #12 .

### DIFF
--- a/src/Qowaiv/Text/Base32.cs
+++ b/src/Qowaiv/Text/Base32.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Text;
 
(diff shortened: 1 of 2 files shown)


erik-am:
@Corniel @plat2059 Found a bug, or at least a change in the contract. Using a `0` (outside of the alphabet) returns a `FormatException` in the original implementation, but doesn't in the new implementation.

Corniel:
That was also the case for the previous implementation. ;)

PR #14, opened 2017-12-19
Make the version part of th UUID

Corniel:
Extended the UUID with the UUID version (4 = random, 3 = MD5 based, 5 = SHA1 based), with according factory methods.

### DIFF
--- a/src/Qowaiv/Uuid.cs
+++ b/src/Qowaiv/Uuid.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Runtime.Serialization;
+using System.Security.Cryptography;
 using System.Text.RegularExpressions;
 using System.Xml;
 using System.Xml.Schema;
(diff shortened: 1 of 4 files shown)


src/Qowaiv/Uuid.cs
```diff
@@ -7,436 +7,435 @@
...
+            Guard.NotNull(writer, "writer");
+            writer.WriteString(ToString(CultureInfo.InvariantCulture));
+        }
+
+        #endregion
+
+        #region (JSON) (De)serialization
+
+        /// <summary>Generates a UUID from a JSON null object representation.</summary>
+        void IJsonSerializable.FromJson()=>m_Value = default(Guid);
```
plat2059: Nitty-gritty mode on: some whitespace around => ?

src/Qowaiv/Uuid.cs
```diff
@@ -7,436 +7,435 @@
...
+                return true;
+            }
+            return false;
+        }
+
+        /// <summary>Generates an <see cref="Uuid"/> applying a <see cref="MD5"/> hash on the data.</summary>
+        public static Uuid GenerateWithMD5(byte[] data)
+        {
+            Guard.NotNull(data, nameof(data));
+            var hash = MD5.Create().ComputeHash(data);
```
plat2059: Shouldn't the instance of MD5 be disposed?

src/Qowaiv/Uuid.cs
```diff
@@ -7,436 +7,435 @@
...
+            UuidExtensions.SetVersion(hash, UuidVersion.MD5);
+            return new Guid(hash);
+        }
+
+        /// <summary>Generates an <see cref="Uuid"/> applying a <see cref="SHA1"/> hash on the data.</summary>
+        public static Uuid GenerateWithSHA1(byte[] data)
+        {
+            Guard.NotNull(data, nameof(data));
+            var bytes = SHA1.Create().ComputeHash(data);
```
plat2059: Same here. Using?

src/Qowaiv/UuidExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+        public static UuidVersion GetVersion(this Guid guid)
+        {
+            var bytes = guid.ToByteArray();
+            var version = bytes[7] >> 4;
+            return (UuidVersion)version;
+        }
+
+        internal static void SetVersion(byte[] uuid, UuidVersion version)
+        {
+            uuid[7] &= 0x0F;
```
plat2059: Doubting here if you should only allow changing the empty version to the passed version.

src/Qowaiv/Uuid.cs
Corniel: It should. 👍

PR #24, opened 2018-08-02
New Factory methods for Qowaiv.ComponentModel.Result

Corniel:
As described in the (updated) README.md

---------------------------------------

### Result model
Also we propose a Result model that includes the validation messages, and Result which can contain both an object and validation messages. This can be a helpful return type for methods that need to return objects but first have to validate them.

``` CSharp
Result<DataType> result = Result.For(data);
```

### ValidationMessages with severity
The difference with Microsoft's default ValidationResult and ValidationMessages is that in this PR we support a severity: info, warning, or error.

Those messages can be created via factory methods:
``` CSharp
var none = ValidationMessage.None;
var info = ValidationMessage.Info(message, args);
var warn = ValidationMessage.Warning(message, args);
var error = ValidationMessage.Error(message, args);
```

Or contained by a Result or Result&lt;T&gt;:

``` CSharp
Result result = Result.WithMessage(messages);
Result<DataType> result = Result.WithMessage<DataType>(messages);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ Also we propose a Result model that includes the validation messages, and Result
 
 ``` CSharp
 Result<DataType> result = Result.For(data);
+Result<DataType> resultWithMessages = Result.For(data, messages);
 ```
 
 ### ValidationMessages with severity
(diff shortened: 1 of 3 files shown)


Corniel:
@Sjaaky Added:

``` Charp
/// <summary>Creates a <see cref="Result{T}"/> for the data.</summary>
public static Result<T> For<T>(T data, params ValidationResult[] messages) => new Result<T>(data, messages);
```

Is that what you had in mind?

PR #28, opened 2018-09-22
Introduction of Qowaiv.DomainModel

Corniel:
Introduction of Qowaiv.DomainModel. Idea is to have a solution for a robust Entity<TId> base class, a ValueObject base class and some useful interfaces and maybe other base classes.

``` CSharp
public class MyEntity : Entity<Guid>
{
    public MyEntity() : base(){}

    [Mandatory, MaxLength(30), Display(Name = "Full name")]
    public string FullName
    {
        get => GetProperty<string>();
        set => SetProperty(value);
    }

    [Mandatory, Display(Name = "Date of birth")]
    public Date DateOfBirth
    {
        get => GetProperty<Date>();
        set => SetProperty(value);
    }
}

var entity = new MyEntity();
entity.SetProperties(()=>
{
    entity.FullName = "Jimi Hendrix";
    entity.Date = new Date(1942, 11, 27);
})
```

It is 'Always Valid' out of the box. Properties are stored in an dictionary keeping track of the validation attributes (and the display attribute).

Probably the 'easiest'  way to guarantee the immutability of a value object (as class) is by adding a roslyn rule to do so. The `Validate()` method (in this example in the constructor). Throws if the `ValueObject` is not valid.

```csharp
public sealed class Address : ValueObject<Address>
{
    public Address(string street, HouseNumber number, PostalCode code, Country country)
    {
        Street = street;
        HouseNumber = number;
        PostalCode = code;
        Country = country;
        Validate();
    }

    [Mandatory]
    public string Street { get; }
    [Mandatory]
    public HouseNumber HouseNumber { get; }
    public PostalCode PostalCode { get; }
    [Mandatory]
    public Country Country { get; }

    public override bool Equals(Address other)
    {
        if (AreSame(other))
        {
            return true;
        }
        return NotNull(other)
            && Street == other.Street
            && HouseNumber == other.HouseNumber
            && PostalCode == other.PostalCode
            && Country == other.Country
            ;
    }

    protected override int Hash()
    {
        return QowaivHash.HashObject(Street)
            ^ QowaivHash.Hash(HouseNumber, 3)
            ^ QowaivHash.Hash(PostalCode, 5)
            ^ QowaivHash.Hash(Country, 13)
            ;
    }
}
```


Corniel:
Although it is easy to keep track of IsDirty, I decided to drop this, as it is not DDD. Hopefully, @erik-am likes this change. :)

Corniel:
A new attempt, taking `Qowaiv.Valdation.Abstractions` into account, in started with #63 .

PR #30, opened 2018-09-23
Month.Days(Year year)

Corniel:
Simple extension to month.


Corniel:
@Sjaaky Do you have a better name for this method? (According to your WatsApp comment, you apparently think it is not clear)

PR #39, opened 2019-03-03
Introduction of the ImmutableAttribute to decorate properties that should not be changed after initialization

Corniel:
It is a prerequisite for #28, where an ID should be decorated as such, but it deserves a change set on its own.


Corniel:
After some consideration, I think will not work.

Issue #48, opened 2019-04-25
Introduce a Yes-no/polar SVO


Corniel:
The answer to yes-no or polar question is obviously yes or no. In DDD models it makes sense to use a specific type, and not a `boolean`.

Things to decide:
* Name of the type (YesNo, Polar, something else)
* formatter chars (lowercase, Capitalized first, UPPERCASE)
* how to cast from to `bool`

See also: https://en.wikipedia.org/wiki/Yes–no_question

PR #51, opened 2019-04-26
Implementation of Yes-no

Corniel:
As described in #48 

Casting from and to `Boolean` still has to be done. Also we could consider adding things like:

``` C#
var answer0 = YesNo.Yes;
var answer1 = YesNo.Unknown;
var and = answer0.And(answer1); // YesNo.Unkown;
var or = answer0.Or(answer1); // YesNo.Yes;
```


Corniel:
@Sjaaky : Do you have a suggestion on how to serialize this (JSON/XML)? "y/n/?", "yes/no/?"?

Issue #32, opened 2018-09-24
parsing email address from email's header


NickyNest:
As usual 'from' email's header contains not only email address itself but sender info as well, e.g. 
"John Doe &#60;john.doe@domain.com&#62;".
Would be nice to parse such kind of emails via EmailAddress.Parse().

Corniel:
@NickyNest Fair point. The question is, however, should it strip Jon Doe <..>, and leave the plain e-mail address or not in your opinion? The idea of this SVO is (in mine opinion at least) a way to store a uniform adress, that is why John.Doe@Domain.COM is also stored as john.doe@domain.com.

From a DDD perspective you could argue, that besides the 'simple' or 'plain' email address, you also want to have one that has a 'labeled' one, that can be constructed from a label (name) and a 'plain'  email address.

So, I'm really curious about your use case, and what would you like to see.

NickyNest:
@Corniel My use case is quite simple: I need to grab only plain email address from parsed email headers, so leaving just a uniform address is enough for my purpose. Adding 'label' might break YAGNI at the moment )

Btw, there is already [MailAddress ](https://docs.microsoft.com/en-us/dotnet/api/system.net.mail.mailaddress.-ctor?redirectedfrom=MSDN&view=netframework-4.7.2#System_Net_Mail_MailAddress__ctor_System_String_) class built in framework which includes all the parts (DisplayName, User, Host, Address) and can parse addresses with user info.

Corniel:
I know of the existance of MailAddress (https://github.com/dotnet/corefx/blob/master/src/System.Net.Mail/src/System/Net/Mail/MailAddress.cs)
But not serializable ect...

If I understand your use case well (And I'm willing to add it, is the following):

```CSharp
var email = EmailAddress.Parse("John Doe <john.doe@domain.com>");
Assert.AreEqual("john.doe@domain.com", email.ToString());
```

Corniel:
See pull request #35

Corniel:
@NickyNest Off-topic: In what kind of application(s) are you using Qowaiv if I may ask? And how did you find it?

NickyNest:
>var email = EmailAddress.Parse("John Doe <john.doe@domain.com>");
> Assert.AreEqual("john.doe@domain.com", email.ToString());
That is exactly what I meant.

[...]

PR #54, opened 2019-05-02
Improve performance of parsing email addresses

Corniel:
Instead of using an (slow, hard to maintain) regular expression, I introduced a dedicated parser to validate and parse an email address.

As a result of that, it was also easy to implement #32 .

@wefrieke : If i use `IPAddress.TryParse()` IP-adresses with leading zero's are valid, while the regex was forbidding such. Also with this change there is support for IP6. Is that what we want?


Corniel:
@wefrieke 

``` C#
IPAddress.TryParse("111.222.333", out ip); // => "111.222.1.177"
```

So, quite forgiving.

Issue #57, opened 2019-07-08
MandatoryAttribute should allow zero value for nullable decimal


LauraXiulan:
The MandatoryAttribute in ComponentModel.DataAnnotations currently does not allow a zero value for nullable decimals. It should be allowed as it is a valid value.

PR #58, opened 2019-07-08
MandatoryAttribute should allow nullable values with default of underlying type

Corniel:
``` C#
public class Model
{
    [Mandatory]
    public decimal? Value { get; set; }
}
```

This model should be valid if `Value` is set to `0`. This is not the case. Reason is that the `IsValid()` method is provided with an object. In other words, with the boxed value. That strips off the nullable wrapper. So instead of comparing with the default of `decimal?` which is a not set `decimal`, it compares with a default of `decimal` which is zero.

To overcome this, I did override `ValidationResult IsValid(object, ValidationContext)`. By doing so, I can (with reflection, that is the downside) get the type of the property. I do not cache that. Reason not to do so is that nothing forbids reusing attributes for different values. Of course, you could cache the outcome for a type with member name combination, but that feels to far fetched.

See issue #57


Sjaaky:
What about the `ValidationContext`?

PR #65, opened 2019-07-25
Added IBAN support for extra countries

Corniel:
Official IBAN countries:
* BY (Belarus)
* TL (Timor Leste)
* VA (Vatican City)

Candidate countries:
* KM (Comoros)
* TD (Chad)
* EG (Egypt)
* GA (Gabon)
* HN (Honduras)
* MA (Morocco)
* NI (Nicaragua)
* NE (Niger)
* TG (Togo)


Corniel:
@pmdevers All strings is stripped first before validated, so that did not add any value.

Issue #68, opened 2019-07-26
Parser not supporting 3 more formats


salmarko:
https://github.com/Qowaiv/Qowaiv/blob/77a99cba7f27fc1e92e63d7efc794138f47e7efe/src/Qowaiv/EmailParser.cs#L25

These 3 formats are not supported at the moment:

'"XX Rotterdam <xx@domain.nl>" <xx@domain.nl>'
'"XX \r\n Rotterdam <administratie@domain.nl>" <administratie@domain.nl>'
'"XX Rotterdam <xx@inkoop.exactonline.nl>" <xx@inkoop.exactonline.nl>'

Can you make changes to support them as well?

Corniel:
In other words: email addresses that have a display name in double quotes? I think I can.

salmarko:
Thx :) 👍

Corniel:
I Have a question though: what escaping character is used to escape the double quote?

salmarko:
This is a part of json:

"date":"Mon, 22 Jul 2019 00:52:24 -0700 (PDT)",
"to":[  
"\"Rozey Rotterdam <rotterdam@rozey.nl>\" <rotterdam@rozey.nl>",
"\"Rozey\r\n Rotterdam <administratie@lacubanita.nl>\" <administratie@lacubanita.nl>",
"\"Rozey Rotterdam <k24f@inkoop.exactonline.nl>\" <k24f@inkoop.exactonline.nl>"
],
"messageId":"<5d356b38.1c69fb81.22d9e.bede@mx.google.com>",
"subject":"Factuur 201912885 van 21-07-2019"


So, \

Corniel:
@salmarko Nide some stuff because I think it is data that should not be published out in the open.

PR #69, opened 2019-07-26
Extend support for display names

Corniel:
Qowaiv supports email addresses with a display name: 
* `Joe Smith <email@qowaiv.org>`
* `(Joe Smith) email@qowaiv.org` (in fact as comment)

The change includes:
* `"Joe Smith" email@qowaiv.org`

Questions to answer: 
* Should also `'Joe Smith' email@qowaiv.org` be supported?
* Which escape character(s) for the quote should be supported, if any?
* Should a space character after the end quote be mandatory?

See: #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,11 @@ case-insensitve.
 
 ``` C#
 var email = EmailAddress.Parse("Test Account <TEST@qowaiv.org>");
+var quoted = EmailAddress.Parse("\"Joe Smith\" email@qowaiv.org");
 var ip_based = EmailAddress.Parse("test@[172.16.254.1]");
 
 email.ToString(); // test@qowaiv.org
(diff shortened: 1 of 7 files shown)


Corniel:
@salmarko Thanks for the approval, but those 3 questions still have to be answered.

salmarko:
_Should also 'Joe Smith' email@qowaiv.org be supported?_ (No, as far as I am concerned)
_Which escape character(s) for the quote should be supported, if any?_ (Only \ )
_Should a space character after the end quote be mandatory?_ (Yes)

Corniel:
@salmarko :
* Added support for '\' escape character.
* Made whitespace character afterward mandatory
* Added `WithDisplayName(string)` method to convert back to something with a display name. (always Display name <emailaddress>)

PR #70, opened 2019-08-08
Introduction of the extension method Percent().

Corniel:
Currently percentages can be created as follows:
``` C#
Percentage p = 0.0314; // implict cast: 3.14%
var p = Percentage.Parse("3.14"); //  Parse: 3.14%;
var p = Percentage.Parse("3.14%"); // Parse: 3.14%;
```
This is convenient (to a high extend) as it makes casting both ways (from and to floating points) predictable, and easy. However, for defining `static readonly`  values in code, it is not ideal.

To overcome this, it might worth to consider the following:

``` C#
var p = 3.14.Percent();
```

### DIFF
--- a/src/Qowaiv/PercentageExtensions.cs
+++ b/src/Qowaiv/PercentageExtensions.cs
@@ -360,5 +360,18 @@ namespace Qowaiv
         public static UInt16 Divide(this UInt16 d, Percentage p) { return (UInt16)((Decimal)d).Divide(p); }
 
         #endregion
+
+        #region Percent()
+
+        /// <summary>Interpertates the <see cref="int"/> if it was written with a '%' sign.</summary>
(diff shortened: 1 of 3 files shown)


src/Qowaiv/PercentageExtensions.cs
```diff
@@ -360,5 +360,18 @@ public static class PercentageExtensions
         public static UInt16 Divide(this UInt16 d, Percentage p) { return (UInt16)((Decimal)d).Divide(p); }
 
         #endregion
+
+        #region Percent()
+
+        /// <summary>Interpertates the <see cref="int"/> if it was written with a '%' sign.</summary>
```
advdh: typo: Interprets

Corniel: Thanks. But do you like the concept?

PR #81, opened 2019-09-10
Support for OpenAPI

Corniel:
- [ ] Which types need a pattern?!
- [ ] Short descriptions
- [ ] Should titles, min- and maxlength and maybe other properties conditionally be supplied?

The [OpenAPI Specification](https://swagger.io/docs/specification/about/)
(formerly Swagger Specification) is an API description format for REST API's.

To improve usage of your REST API's you should specify the Data Type of your
SVO's. To make this as simple as possible, Qowaiv SVO's are decorated with the
`OpenApiDataTypeAttribute`. It specifies the type, format, (regex) pattern,
and if the data type is nullable, all when applicable.

| SVO                                      | Type    | Format              | Nullable |
|------------------------------------------|---------|---------------------|----------|
| Date                                     | string  | date                | false    |
| EmailAddress                             | string  | email               | true     |
| EmailAddressCollection                   | string  | email-collection    | true     |
| Gender                                   | string  | gender              | true     |
| HouseNumber                              | string  | house-number        | true     |
| LocalDateTime                            | string  | date-time-local     | false    |
| Month                                    | string  | month               | true     |
| Percentage                               | string  | percentage          | false    |
| PostalCode                               | string  | postal-code         | true     |
| Uuid                                     | string  | uuid-base64         | true     |
| WeekDate                                 | string  | date-weekbased      | false    |
| Year                                     | integer | year                | true     |
| YesNo                                    | string  | yes-no              | true     |
| Financial.Amount                         | number  | amount              | false    |
| Financial.BankIdentifierCode             | string  | bic                 | true     |
| Financial.Currency                       | string  | currency            | true     |
| Financial.InternationalBankAccountNumber | string  | iban                | true     |
| Financial.Money                          | string  | money               | false    |
| Globalization.Country                    | string  | country             | true     |
| IO.StreamSize                            | integer | stream-size         | false    |
| Security.Cryptography.CryptographicSeed  | string  | cryptographic-seed  | true     |
| Statistics.Elo                           | number  | elo                 | false    |
| Web.InternetMediaType                    | string  | internet-media-type | true     |


Corniel:
@mvmaasakkers : But, which types need a pattern? What should be in the descriptions?

Corniel:
@pmdevers @mvmaasakkers : As no patterns have been specified yet, do you think those are not needed? I'm in doubt. You could argue, that given the format (such as email, percentage and such), the format is clear enough by itself, and does not require to specify the regex pattern too, but the API might be consumed by code that does not have explicit value types... #indoubt

mvmaasakkers:
Email is clear to me, just a string according to [RFC 5322](https://tools.ietf.org/html/rfc5322). But percentage is not... is it a string ("1.32%") or a decimal (1.32)? So that would need a specific description. Same as for example the string month. I assume there are only 12 options there but which ones are they (localized or not, capitalized or not?). Globalization.Country, is that [ISO 3166 Alpha-3](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes)?

The moment you won't accept a free format string through json more information is needed. You could always let your validation return the issues for you but defining it upfront would enable developers to implement it more user friendly.

Corniel:
@mvmaasakkers I agree that it needs clarification, that is beyond discussion. But should it be a description (only), or a pattern that makes it as clear as possible? Or should it be both?

Percentage as an example:
``` JavaScript
{
    type: string,
    format: percentage,
    pattern: '^-?[0-9]+(\.[0-9]+)?%$',
    description: 'ratio expressed as a fraction of 100 denoted using the percent sign "%".'
}
```

Issue #87, opened 2019-09-11
BIC means Business Identifier Code not Bank Identifier Code


Corniel:
See https://en.wikipedia.org/wiki/ISO_9362

PR #89, opened 2019-09-16
Introduction of BusinessIdentifierCode

Corniel:
BIC means Business Identifier Code not Bank Identifier Code. So the Single Value Object should be named `BusinessIdentifierCode` not `BankIdentifierCode`. The old SVO has been deprecated, and the new one been introduced.

To consider: remove the LocationCode, BusinessCode, CountryCode, and BranchCode sufix. (not breaking as new type).

See #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Represents a BIC as specified in ISO 13616.
 var bic = BusinessIdentifierCode.Parse("AEGONL2UXXX");
 
 var business = bic.Business; // "AEGO"
-var country = bic.Counrty; // Country.NL
+var country = bic.Country; // Country.NL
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -147,9 +147,19 @@ A seed, representing random data to encrypt and decrypt data.
...
 
-### Bank Identifier Code (BIC)
+### Business Identifier Code (BIC)
 Represents a BIC as specified in ISO 13616.
 
+``` C#
+var bic = BusinessIdentifierCode.Parse("AEGONL2UXXX");
+
+var business = bic.Business; // "AEGO"
+var country = bic.County; // Country.NL
```
LauraXiulan: Country

PR #95, opened 2019-10-12
Introduction of a Date span

Corniel:
A date span represents duration between to dates. As opposed to `System.TimeSpan`, a data span has not sub-day precision, and has notation of months too. This is typically useful when calculating ages, (monthly) subscriptions and such.

My suggestion would be to have to have the following serialisation:
``` C#
var span = new DateSpan(years: 3, months: 2, days: -4);
Assert.AreEqual("+3Y+2M-4D", span.ToString());
```
I'm not sure (yet) if the different blocks (years, months, days) should be mandatory, and `DateSpan.Zero` seralizes as `0D` now.


src/Qowaiv/LocalDateTime.cs
```diff
@@ -229,6 +229,47 @@ public LocalDateTime Add(TimeSpan value)
             return new LocalDateTime(this.Ticks + value.Ticks);
```
pmdevers: this,Ticks

Moet dit niet 'm_Value.Ticks' zijn

Issue #84, opened 2019-09-11
Month is JSON serialized as numeric string


Corniel:
When serializing a `Month` with JSON (and with XML too), It should use the invariant name, short or long, that is a debate, but not the numeric value (as string).

PR #113, opened 2019-12-03
Split op SVO struct files

Corniel:
To make a clear distinction between code that should and that should not be added after (initial) generation, a split-up of all SVO's should be done.

This should also help in showing people where "_the magic is happening_".


Corniel:
By doing this I (more or less by change ;)) fixed #84 ,

PR #115, opened 2019-12-05
Simplify JSON serialization

Corniel:
When Qowaiv was founded, JSON was still relatively new. I decided to support that via a interface comparable to the `IXmlSerializable` interface:

``` C#
namespace Qowaiv.Json
{
    public interface IJsonSerializable
    {
        void FromJson();
        void FromJson(string jsonString);
        void FromJson(long jsonInteger);
        void FromJson(double jsonNumber);
        void FromJson(DateTime jsonDate);
        object ToJson();
    }
}
```

It was mapping more or less 1-to-1 to Newtonsoft's JSON serializers and was working okay. However, it has some downsides:
1. You have to explicitly specify with JSON nodes you do not support.
2. `DateTime` is not JSON node type, it is a Newtonsoft feature.
3. `Boolean` is missing.
4. `FromJson()` should not be called, it should be dealt with by the Serializer.
5. Like the `IXmlSerializable` interface, for deserialization, the state of the SVO is alternated afterward, which can lead to breaking the read-only contract.

In this pull request, usage of the interface has been removed completely and the `IJsonSerializable` has been marked as `[Obsolete(IsError:true)]`. All SVO's that had the interface, now have at least one static `FromJson(json)` factory method (excepting a `string` as input), and one `ToJson()` method, that in most cases returns a `string`.

I'm working on providing two separate packages that ship a Newtonsoft and System.Text.Json based JSON serializer that can deal with this convention based SVO's: https://github.com/Qowaiv/qowaiv-json/pull/1


src/Qowaiv/EmailAddressCollection.cs
```diff
@@ -253,63 +253,20 @@ protected virtual void WriteXml(XmlWriter writer)
...
+        public static EmailAddressCollection FromJson(string json) => Parse(json, CultureInfo.InvariantCulture);
 
-        /// <summary>Converts an email address into its JSON object representation.</summary>
-        object IJsonSerializable.ToJson() { return ToJson(); }
-        /// <summary>Converts an email address into its JSON object representation.</summary>
-        /// <remarks>
-        /// this is used by IJsonSerializable.FromJson() so that it can be changed by derived classes.
-        /// </remarks>
-        protected virtual object ToJson() => Count == 0 ? null : ToString(CultureInfo.InvariantCulture);
+        /// <summary>Serializes the date to a JSON node.</summary>
```
LauraXiulan: date?

src/Qowaiv/Financial/Currency.cs
```diff
@@ -116,36 +116,20 @@ public IEnumerable<Country> GetCountries(Date measurement)
...
-        /// </param>
-        void IJsonSerializable.FromJson(DateTime jsonDate) => throw new NotSupportedException(QowaivMessages.JsonSerialization_DateTimeNotSupported);
+        /// <returns>
+        /// The deserialized currency.
+        /// </returns>
+        public static Currency FromJson(long json) => FromJson(json.ToString("000", CultureInfo.InvariantCulture));
 
-        /// <summary>Converts a currency into its JSON object representation.</summary>
-        object IJsonSerializable.ToJson() => m_Value == default ? null : ToString(CultureInfo.InvariantCulture);
+        /// <summary>Serializes the date to a JSON node.</summary>
```
LauraXiulan: date?

PR #123, opened 2020-02-20
Introduction of Uuid.NewSequential()

Corniel:
As UUID's are commonly used for the clustered key of a database table. For
massive database with a lot of inserts (they go hand in hand normally) this can
be a performance issue, as by default generated UUID's are not sequential, so
the clustered index gets a lot of random inserts.

By using a sequential UUID this problem can be minimized. Obviously, if you can
fully rely on the sequential UUID generation by your database of choice, you
should consider that, but in most cases you want to generate the ID upfront.
In that case `Uuid.NewSequential()` comes handy:

``` C#
var uuid = Uuid.NewSequential(UuidComparer.SqlServer);
```

As databases might (like SQL Server does) order your UUID/GUID's differently
that .NET does, this generator does that too. Also keep in mind that this generated ID
is not perfectly sequential; first of all because it has a 0.32 nanosecond
overlap, but more seriously, as some time may elapse between the generation and
the storage in the database. Furthermore, these generated UUID's are not sequential
once mixed with the sequential generated UUID's by your database.


Corniel:
@advdh I'm in doubt about the filled part. If I have a massive database table (and clustered index) already, does it done matters if I insert (more or less) at the end, or is it equally fine to insert it halfway, as long it is always more or less after the previous insert?

advdh:
It's better to append new records really at the end. Inserting batches half way is less optimal because it causes page splits, and that leads to fragmentation. By the way: in eol primary keys (guids) are usually not the clustered index, because it's more important to have the division (tenantId) as first clustered index column.

Corniel:
@advdh Well, in that case, I keep the _filled_ option out, as it would be extremely hard (if not impossible), to come up with a generic solution that works.

PR #128, opened 2020-02-24
Introduction of Qowaiv.Mathematics.Fraction

Corniel:
It was about time to add a representation of a fraction:

``` C#
Fraction fraction = 4.DividedBy(13); // new Fraction(4, 13)
Fraction parsed = Fraction.Parse("4/13");
Fraction parsed = Fraction.Parse("4½"); // single fraction character
Fraction parsed = Fraction.Parse("4²³⁄₄₇"); // unicode super- and subscript
Fraction created = Fraction.Create(4.0/13.0);

Fraction zero = default(Fraction); // Fraction.Zero
```

As the default value of a struct can always be created, is should have a meaningful value. There are basically two options:
1. Undefined/NaN (as 0/0 well, you get the point)
2. Zero (as that makes sense)

If chosen to go for option 2, as going for option would 1 require a lot of exception throwing, as most operations on NaN do not make sense.

Given option 2 you have 2 options:
1. Treat Zero as a special case.
2. make the 'real' `Denominator` => `denominator + 1`, so that default would effectually be 0/1.

Given that zero has a special meaning (a + 0 = a, a * 0 = 0), I decided to go for 1 (for now).


src/Qowaiv/Mathematics/FractionParser.cs
```diff
@@ -0,0 +1,303 @@
...
+                    {
+                        return null;
+                    }
+                    tokens |= Tokens.Space;
+                    tokens &= Tokens.Reset;
+                }
+                else if (ch.IsVulgar(out var vulgar))
+                {
+                    // A vulgar is only allowed directly followed by an integer, or as single character.
+                    if (tokens.HasAny(Tokens.Operator | Tokens.SuperScript| Tokens.SubScript) || !str.EndOfBuffer(index))
```
Corniel: Rename to Bar

src/Qowaiv/Mathematics/Fraction.cs
```diff
@@ -0,0 +1,783 @@
...
+                }
+                // use invariant as we want to convert to superscript.
+                var super = remainder.Abs().ToString(CultureInfo.InvariantCulture).Select(ch => Formatting.SuperScript[ch - '0']).ToArray();
+                sb.Append(super);
+            }
+            else
+            {
+                if (sb.Length != 0)
+                {
+                    sb.Append(' ');
```
Corniel: And last caracter is not ' '.

PR #132, opened 2020-03-05
Support for strongly-typed Identifiers

Corniel:
Specially with identifiers (GUID, long, string, doesn't really matter) the risk of [primitive obsession](https://andrewlock.net/using-strongly-typed-entity-ids-to-avoid-primitive-obsession-part-1/) is substantial.

There are multiple ways to overcome this:

1. Ad-hoc implementation when needed.
2. Using a base class (one per underlying type).
3. Generated code.
4. The approach of this pull request (🙂).

All have there downsides:
1. A lot of work, not feature complete, a lot of code duplication.
2. Null ID's, static contracts can not be inherit so still a lot of code.
3. Adds (a lot of) complexity to the build.
4. Generics.

So what does this pull-request/approach try to achieve?

``` C#
public sealed class ForUser : GuidLogic { }
public sealed class ForPerson : GuidLogic { }

var user = Id<ForUser>.Parse("0bb59085-9184-4df9-b9d4-08e1ba65cef8");
var person = Id<ForPerson>.Parse("0bb59085-9184-4df9-b9d4-08e1ba65cef8");

var same = user.Equals(person); // false

var guid = (Guid)user;

public sealed class ForRecord: LongIdLogic { }
var id = Id<ForRecord>.Parse("007");
var num = (long)id;
```

The trick applied here is that the provided generic must implement `IIdentifierLogic`. For every generic type that is used, a static instance of the `IIdentifierLogic` is used to do the identifier specific logic; so by defining the generic type you both have the unique type per identifier (the starting point) as the option to have (limited but sufficient) dependency injection.

Obviously, the prefix 'For' is optional, but I think is a nice convention. Not in the last place, as in most cases without the prefix, you would have to specify namespaces, as `Id<ForUser>` probably is the identifier of some `User` class.


src/Qowaiv/Identifiers/Id.cs
```diff
@@ -0,0 +1,398 @@
...
+        /// <returns>
+        /// The serialized JSON node.
+        /// </returns>
+        public object ToJson() => IsEmpty() ? null : logic.ToJson(m_Value);
+
+        private string CastToString()
+        {
+            if (logic.BaseType != typeof(string))
+            {
+                throw Exceptions.InvalidCast(logic.BaseType, typeof(string));
```
pmdevers: or GetTyp()

Corniel:
@Sjaaky & @pmdevers What do you think guys? object as internal, or a byte[]?

pmdevers:
@Corniel what is the purpose of this intenal object of byte[]?

Corniel:
@pmdevers Well, it is the underlying value. By using a byte[] we need less 'magic'. the Identifier logic as we don't have to cast (and box/unbox). The downside is that we create a byte array for int/long/guid/string and that is overhead.

src/Qowaiv/Reflection/QowaivType.cs
```diff
@@ -89,5 +91,99 @@ public static Type GetNotNullableType(Type objectType)
...
+            {
+                return sb.AppendType(type.GetElementType(), withNamespace).Append("[]");
+            }
+
+            if (primitives.TryGetValue(type, out var primitive))
+            {
+                return sb.Append(primitive);
+            }
+            
+            if (type.IsGenericTypeDefinition)
```
Corniel: Add IsNested logic

src/Qowaiv/Identifiers/Id.cs
```diff
@@ -0,0 +1,382 @@
...
+                ? Empty
+                : new Id<TIdentifier>(logic.FromBytes(bytes));
+        }
+
+        /// <summary>Creates an identifier from an <see cref="object"/>.</summary>
+        /// <param name="obj">
+        /// The <see cref="object"/> to create an indentifier from.
+        /// </param>
+        /// <exception cref="InvalidCastException">
+        /// if the identifier could not be created form the <see cref="object"/>.
```
LauraXiulan: f**ro**m

src/Qowaiv/Identifiers/GuidLogic.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// 
+        /// P
+        /// 32 digits separated by hyphens, enclosed in parentheses:
+        /// (00000000-0000-0000-0000-000000000000)
+        /// 
+        /// X
+        /// Four hexadecimal values enclosed in braces, where the fourth value is a subset of eight hexadecimal values that is also enclosed in braces:
+        /// {0x00000000,0x0000,0x0000,{0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00}}
+        /// 
+        /// the lowercase formats are lowercase (except the the 's').
```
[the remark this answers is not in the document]
Corniel: Formatting options on a GUID. have reformatted it.

Corniel:
@pmdevers , @Sjaaky @LauraXiulan 
If you have to choose:
1. IIdentifierLogic
2. IIdentivierBehavior (Ore Behaviour, but I assume USA english should be the standard here)

What would make most sense?

pmdevers:
I would go for **IdentifierBehavior**

PR #139, opened 2020-03-18
Introdcution of Type.ToCSharpString()

Corniel:
The `ToString()` of generic types is not that convenient. The CSharp format is.

A requirement for strongly typed-id's (#132 ).

### DIFF
--- a/src/Qowaiv/Reflection/QowaivType.cs
+++ b/src/Qowaiv/Reflection/QowaivType.cs
@@ -143,7 +143,7 @@ namespace Qowaiv.Reflection
                 // special case for nullables.
                 if (arguments.Length == 1 && type.GetGenericTypeDefinition() == typeof(Nullable<>))
                 {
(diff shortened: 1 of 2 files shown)


src/Qowaiv/Reflection/QowaivType.cs
```diff
@@ -89,5 +91,108 @@ public static Type GetNotNullableType(Type objectType)
...
+            }
+
+            if (type.IsGenericType)
+            {
+                var arguments = type.GetGenericArguments();
+
+                // special case for nullables.
+                if (arguments.Length == 1 && type.GetGenericTypeDefinition() == typeof(Nullable<>))
+                {
+                    return sb.AppendType(arguments[0], withNamespace).Append("?");
```
LauraXiulan: Single quotes, om het consistent te houden.

Corniel: Done. BTW, English please. :)

PR #140, opened 2020-03-23
Improve ToString(string, IFormatProvider) comment for the `format` parameter

Corniel:
The ToString(string, IFormatProvider) comment for the `format` parameter is wrong.

For `System.DateTime`, this is the text:  A standard or custom date and time format string.


Corniel:
@LauraXiulan : No better suggestion than just remove the "this"?

LauraXiulan:
@Corniel Ah, should I have made a suggestion?

Corniel:
@LauraXiulan Should, should... If you have something better than my 'fix', that would be appreciated. :)

PR #141, opened 2020-03-24
Added YesOrNo() to YesNo

Corniel:
A small but useful addition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,22 @@ has an "empty"(unset) and "unknown" value. It maps easily with a `boolean`, but
 Supports all kind of formatting (and both empty and unknown) that can not be
 achieved when modeling a property as `bool` instead of an `YesNo`.
 
+``` C#
+// Creation
+var yn = YesNo.Parse("ja", new CultureInfo("nl-NL")); // Parsing
+var answer = YesNo.Yes; // costants
+
+if(answer.IsEmptyOrUnknown())
+{
+   // Do logic.
+}
+
+if((bool)answer) // Equal to answer.IsYes()
+{
+    // Do logic.
+}
+```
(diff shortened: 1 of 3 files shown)


test/Qowaiv.UnitTests/YesNoTest.cs
```diff
@@ -716,13 +725,35 @@ public void Explicit_BoolToYesNo_Yes()
...
+        public void Explicit_YesNoToBool(bool expected, YesNo yesNo)
+        {
+            var casted = (bool)yesNo;
+            Assert.AreEqual(expected, casted);
+        }
+
+        [Test]
+        public void Implicit_InIfStatement()
+        {
+            Assert.Throws<DivideByZeroException>(() =>
```
LauraXiulan: Why a divideByZeroException?

README.md
```diff
@@ -168,6 +168,22 @@ has an "empty"(unset) and "unknown" value. It maps easily with a `boolean`, but
 Supports all kind of formatting (and both empty and unknown) that can not be
 achieved when modeling a property as `bool` instead of an `YesNo`.
 
+``` C#
+// Creation
+var yn = YesNo.Parse("ja", new CultureInfo("nl-NL")); // Parsing
+var answer = YesNo.Yes; // costants
```
LauraXiulan: co**n**stants

test/Qowaiv.UnitTests/YesNoTest.cs
Corniel: It does not really matter, but that is one you don't expect to be thrown by an invalid cast, And you want to be sure, that is reaches the block within the if-statement.

PR #143, opened 2020-03-25
Found FP's and FN's according to github/jstedfast/EmailValidation

Corniel:
Investigate differences and fix bugs.

See: https://github.com/jstedfast/EmailValidation


Corniel:
@wefrieke : https://github.com/jstedfast/EmailValidation Claims that
``` C#
 [TestCase("punycode-numbers-in-tld@sld.xn--3e0b707e")]
  ```
Should be a valid email-address, what do you think? (It is now rejected base on the "--".

PR #152, opened 2020-09-15
Implementation Telephone number

Corniel:
Suport of telephone numbers.


src/Qowaiv/TelephoneNumber.cs
```diff
@@ -0,0 +1,138 @@
...
+                }
+                else if (IsUnknown())
+                {
+                    return TelephoneNumberType.Unknown;
+                }
+                else if (m_Value.StartsWith("+", StringComparison.InvariantCulture))
+                {
+                    return TelephoneNumberType.International;
+                }
+                else if (m_Value.StartsWith("0", StringComparison.InvariantCulture))
```
LauraXiulan, line 54: If a telephone number starts with "00" it can also be an international number right, so shouldn't this check if the second is also a 0 or not and return International if so?

test/Qowaiv.UnitTests/TelephoneNumberTest.cs
```diff
@@ -0,0 +1,632 @@
...
+        /// <summary>TelephoneNumber.IsUnknown() should be false for the TestStruct.</summary>
+        [Test]
+        public void IsUnknown_TestStruct_IsFalse()
+        {
+            Assert.IsFalse(TestStruct.IsUnknown());
+        }
+
+        /// <summary>TelephoneNumber.IsEmptyOrUnknown() should be true for the default of telephone number.</summary>
+        [Test]
+        public void IsEmptyOrUnknown_Default_IsFalse()
```
LauraXiulan, line 70: IsTrue

src/Qowaiv/TelephoneNumber.cs
Corniel: 00 can only be international, and will resolve into +.

test/Qowaiv.UnitTests/TelephoneNumberTest.cs
Corniel: Decided not to continue with this for now.

PR #157, opened 2020-11-11
Add IBAN patterns for missing countries

Corniel:
[Wikipedia](https://en.wikipedia.org/wiki/International_Bank_Account_Number#IBAN_formats_by_country) mentions 103 countries that support IBAN, and Qowaiv only has 94. These should been added.
* Djibouti
* Equatorial Guinea
* Guinea-Bissau
* Iraq
* Lybia
* Saint Lucia
* Seychelles
* Sao Tome And Principe
* El Salvador

As there is a standardized way of describing the IBAN pattern, the so called BBAN format, a BBAN to `Regex` method have been added. I also supports literals (`[.+]`) for adding values that fixed, and an option for fixed checksum values, as some counties have those (and it is not part of the BBAN format).


test/Qowaiv.UnitTests/Financial/IBAN_specs.cs
```diff
@@ -0,0 +1,313 @@
...
+                Country.VG,
+                Country.XK,
+            };
+
+            Assert.AreEqual(supported, InternationalBankAccountNumber.Supported.OrderBy(c => c.IsoAlpha2Code));
+            Assert.AreEqual(103, InternationalBankAccountNumber.Supported.Count);
+        }
+    }
+
+    public class Inut_is_invalid_when
```
LauraXiulan, line 125: Input?

test/Qowaiv.UnitTests/Financial/IBAN_specs.cs
```diff
@@ -0,0 +1,313 @@
...
+        public void country_does_not_support_IBAN()
+        {
+            Assert.IsFalse(InternationalBankAccountNumber.IsValid("US20INGB0001234567"));
+        }
+    }
+
+    public class Input_is_valid
+    {
+        [Test]
+        public void for_unkown_IBANs()
```
LauraXiulan, line 173: unknown

src/Qowaiv/Financial/InternationalBankAccountNumber.Bban.cs
```diff
@@ -0,0 +1,47 @@
...
+    public partial struct InternationalBankAccountNumber
+    {
+        /// <summary>Constructs a <see cref="Regex"/> based on its BBAN pattern.</summary>
+        private static KeyValuePair<Country, Regex> Bban(Country country, string bban, int? checksum = default)
+        {
+            var blocks = bban.Split(',').Select(b => b.Buffer());
+            var pattern = CharBuffer.Empty(64)
+                .Add('^')
+                .Add(country.IsoAlpha2Code)
+                .Add(checksum.HasValue ? checksum.Value.ToString("00") : "[0-9][0-9]");
```
LauraXiulan, line 27: Why not use [0-9]{2} instead of [0-9][0-9]?

LauraXiulan (approved):

src/Qowaiv/Financial/InternationalBankAccountNumber.Bban.cs
Corniel: @LauraXiulan also possible. I think this is easier to read. But it is matter of taste.

PR #205, opened 2021-11-04
GetHashcode()'s should be non-predicitable

Corniel:
To support hashing (`object.GetHashCode()`) the hash code should always return the same value, for the same object. As SVO's are equal by value, the hash is calculated based on the underlying value.

For security measures, however, this is only true within the same app domain. By having different hashes for the same value for different app domains, there is good defense against hash flooding.

Nice long reed [andrewlock.net: Why is string.GetHashCode() different each time I run my program in .NET Core?](https://andrewlock.net/why-is-string-gethashcode-different-each-time-i-run-my-program-in-net-core)

To check if the hashes are as expected:
``` C#
using(Hash.WithFixedRandomizer())
{
    var hash = Hash.Code("QOWAIV string");
    hash.Should().Be(1211348473);
}
```
Obviously, this should only be used within test code.


Sjaaky (approved):
Didn't know hashflooding was a concern, but this seems a good way to counter it.

Issue #226, opened 2021-12-22
Percentage can not be initialized on environments lacking French or Iranian culture


krakkepit:
Now that the Percentage class parses to French and Iranian Cultures when no FormatProvider is given, simple docker images will crash because those ICUs are not installed
 
StackTrace
```
System.InvalidOperationException: Failed to convert configuration value at '<REDACTED>' to type 'Qowaiv.Percentage'.
 ---> System.TypeInitializationException: The type initializer for 'Qowaiv.Percentage' threw an exception.
 ---> System.Globalization.CultureNotFoundException: Only the invariant culture is supported in globalization-invariant mode. See https://aka.ms/GlobalizationInvariantMode for more information. (Parameter 'name')
fr-FR is an invalid culture identifier.
   at System.Globalization.CultureInfo..ctor(String name, Boolean useUserOverride)
   at System.Globalization.CultureInfo..ctor(String name)
   at Qowaiv.Percentage..cctor()
   --- End of inner exception stack trace ---
   at Qowaiv.Percentage.TryParse(String s, IFormatProvider formatProvider, Percentage& result)
   at Qowaiv.Percentage.Parse(String s, IFormatProvider formatProvider)
   at Qowaiv.Conversion.PercentageTypeConverter.FromString(String str, CultureInfo culture)
   at Qowaiv.Conversion.SvoTypeConverter`1.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
```

PR #227, opened 2021-12-22
Percentage can not be initialized on environments lacking French or Iranian culture

Corniel:
As reported #226 

By selection the name of the culture provided (if any), this issue can easily be resolved.


krakkepit (approved):
It seems like this PR fixes my issue, since if the culture is not available it will now default to a format instead of showing an error. Thanks!

PR #235, opened 2022-01-15
Percentage.MaxValue should be able to be represented as a string

Corniel:
To represent a percentage as a string its internal `decimal` value must support a multiplication with 10,000 (when using the per ten thousand symbol). Therefor, the `Percentage.MaxValue` and the `Percentage.MinValue` should be reduced by the same factor.


krakkepit (approved):
Seems like a valid fix for the issue at hand. 

I was doubting if there was an easier way to make this work again, since there is probably a property that defines the maximum "stringify"-ible decimal value, but could not find a better way. 

Maybe it is something that we might want to improve on in a later stage.

PR #237, opened 2022-03-22
Fix tests

Corniel:
Two tests where failing due to an invalid setup.


Corniel:
Note to self: investigate lack of coverage.

Sjaaky (approved):

PR #240, opened 2022-05-19
Added HTTP response status Value Object

Corniel:
A Value Object to combine the HTTP response code and the HTTP response reason phrase.

It should not require JSON or XML serialization, and I'm in doubt about `IFormatable`.


Corniel:
After some consideration, abandoned.

Issue #241, opened 2022-05-29
Why are the structs not readonly?


sonnemaf:
Hi,

Is there a reason why the structs are not readonly? It seems that all fields are readonly so they could be marked readonly. This would avoid defensive copies when used in `readonly fields`, `ref readonly locals` and `in` parameters. The primitives of .NET 3+ are (almost) all readonly. I think these should be too.

```cs
readonly struct Percentage {
    ...
}
```

Corniel:
The reason is in the support of `IXmlSerialization`. As Microsoft could work around it for its own primitives (in relation to XML serialization) but we normal mortals can not. Unless, I've missed something.

sonnemaf:
The primitives (int, double, decimal, string) of Microsoft don't implement this interface. Why do you? Maybe you should drop it and make the structs readonly.

At least make all (pure) methods readonly. This will avoid the defensive copies too.

Corniel:
@sonnemaf The don't have to, as they are all special cases in the XML Serializer; they can be found here: [https://source.dot.net/#System.Private.Xml/System/Xml/Serialization/XmlSerializer.cs](https://source.dot.net/#System.Private.Xml/System/Xml/Serialization/XmlSerializer.cs,b9604a3aca6f8489).

I don't understand how I could make [Pure] methods read-only? What do you mean with that?

sonnemaf:
@Corniel I reviewed it and it looks ok. 

I really would have preferred real readonly structs. They show there intent. Noda time had the same issue, see [IMPLEMENTING IXMLSERIALIZABLE IN READONLY STRUCTS](https://codeblog.jonskeet.uk/2018/03/02/implementing-ixmlserializable-in-readonly-structs/)

Maybe it is time for an [C# 8.0 and C# 9.0 training](https://reflectionit.nl/training/csharp/csharp-8-9) 😊

sonnemaf:
Hi @Corniel 

I think I found a solution for the `IXmlSerialization` interface. You can make the structs readonly when you use the `Unsafe.AsRef()` method in the `ReadXml()` method. Have a look at this example using a simple Point struct. The project doesn't need the Unsafe compiler flag. I 'stole' the solution from [NodaTime](https://github.com/nodatime/nodatime/blob/main/src/NodaTime/Interval.cs).

```cs
using System;
using System.IO;
using System.Runtime.CompilerServices;
using System.Xml;
using System.Xml.Schema;
using System.Xml.Serialization;

[...]

PR #243, opened 2022-05-30
Define properties and methods of SVO's as readonly

Corniel:
To prevent defensive copying on the stack of SVO's, all relevant properties and methods (all that do not involve XML serialization) are defined as `readonly`. Thanks for the suggestion @sonnemaf! (see #241)


sonnemaf (approved):
This solution looks ok. I still think that making the structs readonly is a better one. But this is an acceptable alternative.

Corniel:
> This solution looks ok. I still think that making the structs readonly is a better one. But this is an acceptable alternative.

As stated (#241), that is just imposable. We can not drop XML serialization support, period. And Microsoft does not allow/support custom structs to be serialized without implementing `IXmlSerializable`.

Corniel:
@sonnemaf I picked up the idea of using `System.Runtime.CompilerServices.Unsafe.AsRef(this)` thanks for the suggestion.

sonnemaf:
This looks great! Thx

sonnemaf (changes requested):
You can make the PackageReference in the csproj file Conditional on the TargetFramework .NET Standard 2.0.

```
<PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.1"
  	Condition="'$(TargetFramework)'=='netstandard2.0'" />
```

PR #245, opened 2022-05-31
Investigate ISpanFormattable

Corniel:
It is far from trivial to support it, and I see no way to avoid double implementations. Specially supporting `ICustomFormatter` is potential challenging. 

The implementation for [Guid](https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/shared/System/Guid.cs) was a bit of an inspiration.


Corniel:
@sonnemaf I gave it a second go. I think I;'ll will slowly add them.

Sjaaky:
Calling StringFormatter.TryApplyCustomFormatter() and copying the formatted string into a span kinda defeats the purpose of Span's, doesn't it?

Corniel:
> Calling StringFormatter.TryApplyCustomFormatter() and copying the formatted string into a span kinda defeats the purpose of Span's, doesn't it?

You don't gain anything from using `TryFormat()` in those scenario's, that's true. But if the behavior should be the same for both `ToString()` and `TryFormat()`.

The thing that I did not find yet, if it is recommended to implement it, if there is nothing to be gained. What could be the case for at least a part of the string based SVO's.

PR #246, opened 2022-06-02
Limit match durations on Regular Expressions 

Corniel:
When using [System.Text.RegularExpressions](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions) to process untrusted input, pass a timeout. A malicious user can provide input to `RegularExpressions` causing a [Denial-of-Service attack](https://www.us-cert.gov/ncas/tips/ST04-015). [ASP.NET](http://ASP.NET) Core framework APIs that use `RegularExpressions` pass a timeout. `Regex`-es used in Qowaiv, should have a (strict) limitation on the computation time allowed.


Corniel:
> Looks ok

I'm a bit in doubt what the best threshold should be. 1 ms feels save. But more strict might be better?!

sonnemaf:
You could maybe make it configurable

Corniel:
> You could maybe make it configurable

That would introduce a lot of complexity, while by design, these Regexes are all limited. The most complex one (email adress) has been refactored out, because it was a security risk.

PR #248, opened 2022-06-05
Introduction of Svo<SvoBehavior> as a generic for string based SVO's

Corniel:
Creating SVO's for the scope of (just) a project, might feel like a burden that is not worth the effort. Although one might start with [code generation](https://github.com/Qowaiv/qowaiv-codegenerator), it still results in quite some lines of code, that need maintenance, and unit test coverage.

To overcome this I reused the idea behind `Id<TIdBehavior>` for (string based), SVO's. As an example

``` C#
global using MyAlphaNumeric = Qowaiv.Svo<ForMyAlphaNumeric>;

public sealed class ForMyAlphaNumeric : SvoBehavior
{
    public override MaxLength => 60;
    public override Regex Pattern => new($"^([A-Z][0-9])+$");
}
```

Defines a (dedicated) SVO, only allowing alpha numeric characters with a maximum length of 60.

It might be worth to put the relevant code in a separate namespace. I'm considering the following opions:
* Qowaiv.Generic (like System.Collections.Generic(
* Qowaiv.Adhoc 
* Qowaiv.Custom
* Qowaiv.Customization
* Qowaiv.Customizations

Feedback would be nice.


src/Qowaiv/SvoBehavior.cs
```diff
@@ -0,0 +1,235 @@
...
+    /// The format provider.
+    /// </param>
+    [Pure]
+    internal string ToString(string? str, string? format, IFormatProvider? formatProvider)
+    {
+        if (StringFormatter.TryApplyCustomFormatter(format, str!, formatProvider, out string formatted))
+        {
+            return formatted;
+        }
+        else if (str is null)
```
LauraXiulan, line 225: You don't want to do this check first? Or is this in case the custom formatter has specific null logic?

src/Qowaiv/SvoBehavior.cs
```diff
@@ -0,0 +1,235 @@
...
+        else if (str == unknown) return FormatUnknown(default, CultureInfo.InvariantCulture);
+        else return str;
+    }
+
+    /// <summary>Serializes the Single Value Object to an XML string.</summary>
+    /// <param name="str">
+    /// The string representing the Single Value Object.
+    /// </param>
+    [Pure]
+    public virtual string? ToXml(string? str)
```
LauraXiulan, line 117: Exactly the same as ToJson?

Corniel: Yes, but it is up to the user/implementer to keep those in sync (or not).

src/Qowaiv/SvoBehavior.cs
Corniel: Yes (some sort of). if an `ICustomFormatter` is provided, it is up to that formatter completely. It might contain logging (I would not recommand doing so), or preferring to return `null`, whatever.

LauraXiulan (approved):

README.Custom.SVO.md
```diff
@@ -0,0 +1,120 @@
+# Custom Single Value Object
+Creating SVO's for the scope of (just) a project, might feel like a burden that
+is not worth the effort. Although Qowaiv has as [code generator](https://github.com/Qowaiv/qowaiv-codegenerator),
+it still results in quite some lines of code (easily 400), that needs
+maintenance, and unit test coverage.
+
+Another way to deal with this, is by providing a generic SVO where the actual
+behavior is injectable. Qowaiv has two flavors: `Id<TIdBehavior>` and
+`Svo<TSvoBehavior>`. In this document we focus on latter.
```
mars888, line 9: > we focus on _the_ latter.

README.Custom.SVO.md
```diff
@@ -0,0 +1,120 @@
...
+If defined, the `Regex` pattern can act as second line of defense:
+
+``` C#
+public sealed class ForMySvo : SvoBehavior
+{
+    public override Regex Pattern => new("^([A-Z][0-9])+$");
+}
+```
+
+In this example by ensuring that all chars representing the SVO are alpha-numeric.
```
mars888, line 60: Would probably move this after line 51 and make it something like:
> As in the following example, where we ensure that all characters that make up the SVO are alphanumeric.

README.Custom.SVO.md
```diff
@@ -0,0 +1,120 @@
...
+
+    public override string Format(string str, string? format, IFormatProvider? formatProvider)
+        => format == "l"
+        ? str.ToLower()
+        : str;
+}
+```
+
+## Limitations
+It might be worth to point out that this approach has its limitations. Defining
```
mars888, line 117: Maybe?:
> It may be worth pointing out that...

Or even?:
> It should be pointed out that ...

mars888 (approved):

LauraXiulan:
I see you have already put the code in the Customization namespace. That or Custom would be the best namespace in my opinion.

RensReus (approved):

PR #252, opened 2022-07-12
Update sum descriptions

renzo-baasdam:
Sum does not throw exception anymore when empty array (#251), update this in the descriptions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 | version                                                                       | package                                                                     |
 |-------------------------------------------------------------------------------|-----------------------------------------------------------------------------|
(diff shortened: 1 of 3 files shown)


Corniel:
Thanx for spotting this. However, These changes itself are not worth pushing a new version. They will be released with a (soon to com) new version. So please, roll back the version update related changes.

Corniel (changes requested):
No version bump.

renzo-baasdam:
Was also wondering whether it was worth to update, removed it.

Corniel (approved):

Issue #256, opened 2022-08-29
Adding or subtracting decimal to an Amount is incorrectly treated as Percentage


Sjaaky:
**Bug**
Currently `(Amount)10m + 5m == (Amount)60` is true, which is unexpected.
**Cause**
5m is implicitly cast from decimal to 500% (of type Qowaiv.Percentage). Adding 500% to amount 10 results in 60.
**Solution**
Replace the implicit cast from decimal to Percentage with an explicit cast.
@Corniel made we aware that this will be a breaking change.

PR #257, opened 2022-08-29
Prevent implicit casting by introducing obsolete operator overloads

Sjaaky:
As described in #256, the implicit cast to a `Percentage` by the compiler to resolve `Amount + decimal` and `Amount - decimal` can be solved by providing defined overloads. By doing so, with an obsolete warning the developer is notified that his code need changes.

With the next breaking change release (not decided yet) the implicit cast to percentage should be dropped, together with these operator overloads.

### DIFF
--- a/specs/Qowaiv.Specs/Obsolete_code.cs
+++ b/specs/Qowaiv.Specs/Obsolete_code.cs
@@ -13,13 +13,13 @@ public class Will_be_dropped
     public void Date_Tomorrow() => Date.Tomorrow.Should().Be(Clock.Tomorrow());
 
     [Test]
-    public void Amount_Addition_with_implicit_cast() => (50.Amount() + 49m).Should().Be(99.Amount());
+    public void Amount_Addition_with_decimal() => (50.Amount() + 49m).Should().Be(99.Amount());
 
     [Test]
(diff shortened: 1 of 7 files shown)


specs/Qowaiv.Specs/Obsolete_code.cs
```diff
@@ -11,6 +11,19 @@ public class Will_be_dropped
 
     [Test]
     public void Date_Tomorrow() => Date.Tomorrow.Should().Be(Clock.Tomorrow());
+
+    [Test]
+    public void Amount_Addition_with_implicit_cast() => (50.Amount() + 49m).Should().Be(99.Amount());
```
Corniel, line 16: with_decimal

specs/Qowaiv.Specs/Obsolete_code.cs
```diff
@@ -11,6 +11,19 @@ public class Will_be_dropped
...
+
+    [Test]
+    public void Amount_Subtraction_with_implicit_cast() => (90.Amount() - 20m).Should().Be(70.Amount());
```
Corniel, line 19: with_decimal

Corniel (changes requested):

Corniel (approved):
LGTM

Corniel:
@Sjaaky Thanks for your contribution!

PR #258, opened 2022-09-02
Introduction of CAS Registry Number

Corniel:
A CAS Registry Number, is a unique numerical identifier assigned by the Chemical Abstracts Service (CAS), US to every chemical substance described in the open scientific literature. It includes all substances described from 1957 through the present, plus some substances from as far back as the early 1800's.


src/Qowaiv/Chemistry/CasRegistryNumber.cs
```diff
@@ -0,0 +1,172 @@
...
+public readonly partial struct CasRegistryNumber : ISerializable, IXmlSerializable, IFormattable, IEquatable<CasRegistryNumber>, IComparable, IComparable<CasRegistryNumber>
+{
+    /// <summary>Represents an empty/not set CAS Registry Number.</summary>
+    public static readonly CasRegistryNumber Empty;
+
+    /// <summary>Represents an unknown (but set) CAS Registry Number.</summary>
+    public static readonly CasRegistryNumber Unknown = new(long.MaxValue);
+
+    /// <summary>Gets the number of characters of CAS Registry Number.</summary>
+    public int Length => IsEmptyOrUnknown() ? 0 : (int)Math.Ceiling(Math.Log10(m_Value));
```
marijeborst, line 29: waarom niet string.length?

Corniel: Multiple reasons. First, the `.ToString()` contains two dashes by default. Secondly, this one should be faster.

src/Qowaiv/Chemistry/CasRegistryNumber.cs
```diff
@@ -0,0 +1,172 @@
...
+        result = default;
+        if (val == 0) return true;
+        else if (val >= 10_000 && val <= 9_999_999_999)
+        {
+            var checksum = val % 10;
+            var buffer = val / 10;
+            var sum = 0L;
+            var factor = 0;
+
+            while (buffer > 0)
```
marijeborst, line 144: Je doet buffer alleen maar delen door 10 toch? Dan wordt het nooit kleiner dan nul?

Corniel: It's a long, so it stays a long. ;)

src/Qowaiv/Chemistry/CasRegistryNumber.cs
```diff
@@ -0,0 +1,172 @@
...
+    /// True if the string was converted successfully, otherwise false.
+    /// </returns>
+    [Pure]
+    public static bool TryParse(string? s, IFormatProvider? formatProvider, out CasRegistryNumber result)
+    {
+        result = default;
+        var str = s.Unify();
+        if (string.IsNullOrEmpty(str))
+        {
+            return true;
```
LauraXiulan, line 111: Why return true if the string is null or empty?

LauraXiulan (approved):

specs/Qowaiv.Specs/Chemistry/CAS_Registry_Number_specs.cs
```diff
@@ -0,0 +1,489 @@
...
+        => CasRegistryNumber.IsValid("garbage").Should().BeFalse();
+
+    [TestCase("10028-14-3")]
+    [TestCase("10028-15-5")]
+    [TestCase("10028-84-5")]
+    [TestCase("10020-14-5")]
+    [TestCase("10078-14-5")]
+    [TestCase("10128-14-5")]
+    [TestCase("32028-14-5")]
+    public void checksum_mismatches(string number)
```
marijeborst, line 84: Should there be testcases for cas numbers with different lengths?

specs/Qowaiv.Specs/Chemistry/CAS_Registry_Number_specs.cs
```diff
@@ -0,0 +1,489 @@
...
+{
+    [TestCase("")]
+    [TestCase("?")]
+    public void has_length_zero_for_empty_and_unknown(CasRegistryNumber svo)
+        => svo.Length.Should().Be(0);
+
+    
+    [TestCase(5, "73–24–5")]
+    [TestCase(7, "7732-18-5")]
+    [TestCase(8, "10028-14-5")]
```
marijeborst, line 15: Testcase with length 10?

marijeborst (approved):

src/Qowaiv/Chemistry/CasRegistryNumber.cs
```diff
@@ -0,0 +1,172 @@
+﻿#pragma warning disable S1210
+// "Equals" and the comparison operators should be overridden when implementing "IComparable"
+// See README.md => Sortable
+
+namespace Qowaiv.Chemistry;
+
+/// <summary>
+/// A CAS Registry Number, is a unique numerical identifier assigned by the
+/// Chemical Abstracts Service (CAS), US to every chemical substance described
```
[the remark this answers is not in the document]
Corniel: See: https://en.wikipedia.org/wiki/CAS_Registry_Number

specs/Qowaiv.Specs/Chemistry/CAS_Registry_Number_specs.cs
Corniel: The point here is that by changing one digit, the valid number was not longer valid.

specs/Qowaiv.Specs/Chemistry/CAS_Registry_Number_specs.cs
Corniel: I chose to use numbers that are have been assigned and are publicly available. I couldn't find a public one that is.

PR #261, opened 2022-11-04
Extend with .NET 7.0 support

Corniel:
Extend the build to publish a .NET 7.0 version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,9 @@
 
 | version                                                                       | package                                                                     |
 |-------------------------------------------------------------------------------|-----------------------------------------------------------------------------|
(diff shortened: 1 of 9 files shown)


src/Qowaiv/Text/Regxx.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Qowaiv.Text;
+
+internal static class Regxx
```
Sjaaky, line 3: The name Regxx could be improved. Since it is internal, it's not a major issue.

Corniel: I'm open for suggestions.

Issue #278, opened 2022-11-20
Invenstigate failing Code Coverage build


Corniel:
Currently the Code Coverage build fails. See: https://github.com/Qowaiv/Qowaiv/actions/runs/3507648499/jobs/5875496321

pmdevers:
This was mentioned in the log file of the build
```
Testhost process for source(s) '/home/runner/work/Qowaiv/Qowaiv/example/Qowaiv.AspNetCore.Mvc.ModelBinding.UnitTests/bin/Debug/net6.0/Qowaiv.AspNetCore.Mvc.ModelBinding.UnitTests.dll' exited with error: You must install or update .NET to run this application.
```

PR #280, opened 2022-11-21
#278 Code coverage build requires .NET 6 to be installed

pmdevers:


pmdevers:
resolves #278

Corniel:
@pmdevers What should there be added to run the unit tests in all configurations? (so .NET 4.8, and .NET 5.0) too?

Is is just a matter of installing those frameworks too, or is there more to be done?

Corniel (changes requested):
I would like to .NET 5.0 added, and or .NET 4.8, or .NET Core 3.1 to test all target frameworks. If the last one is the easiest to achieve, the target framework of the specs project obviously has to be changed.

.github/workflows/build-test.yaml
```diff
@@ -12,6 +12,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+      - name: Setup .NET 6.0
```
Corniel, line 15:
```suggestion
  with:
          dotnet-version: '3.1.x'
  with:
          dotnet-version: '5.0.x'
```

Corniel:
You decided to go for a matrix in the end? What is your reasoning? (just curious)

Please also update the target frameworks in the specs project to:
``` XML
<TargetFrameworks>netcoreapp3.1;net5.0;net6.0;net7.0</TargetFrameworks>
```

pmdevers:
I have gone with the matrix because of easy of configuration.

Corniel (approved):
Thanx!

Issue #273, opened 2022-11-18
Rename  IFormatProvider formatProvider argument in TryParse


Corniel:
With the added support of `System.IParseble` an issue is raised as all current `IFormatProvider` parameters have the name `formatProvider`, where the contract defines it as `provider`. As this might a breaking change (with named parameters), it should be changed when changing the major.

PR #281, opened 2022-11-21
Fix formatProvider warnings for .NET 7

pmdevers:


Corniel:
See: #273

Corniel (changes requested):
This is blocked until we upgrade to version 7, as renaming this is a breaking change.

Corniel:
Continuation in #360.

PR #285, opened 2023-01-03
Regexes should be CultureInvariant for Qowaiv

Corniel:
All regular expressions in Qowaiv check for alphanumeric input. All this input is within the ASCII range. It should not take different alphabets or numeric systems into account.

### DIFF
--- a/src/Qowaiv/Text/RegOptions.cs
+++ b/src/Qowaiv/Text/RegOptions.cs
@@ -17,9 +17,9 @@ internal static class RegOptions
 #if NET7_0_OR_GREATER
     public const RegexOptions Default = RegexOptions.Compiled | RegexOptions.NonBacktracking;
 #else
-    public const RegexOptions Default = RegexOptions.Compiled;
+    public const RegexOptions Default = RegexOptions.Compiled | RegexOptions.CultureInvariant;
 #endif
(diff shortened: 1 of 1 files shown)


src/Qowaiv/Text/RegOptions.cs
```diff
@@ -17,9 +17,9 @@ internal static class RegOptions
 #if NET7_0_OR_GREATER
     public const RegexOptions Default = RegexOptions.Compiled | RegexOptions.NonBacktracking;
 #else
-    public const RegexOptions Default = RegexOptions.Compiled;
+    public const RegexOptions Default = RegexOptions.Compiled | RegexOptions.CultureInvariant;
 #endif
     public const RegexOptions IgnoreCase = Default | RegexOptions.IgnoreCase;
     public const RegexOptions RightToLeft = RegexOptions.Compiled | RegexOptions.RightToLeft;
```
LauraXiulan, line 23: Do you want the CultureInvariant here as well?

LauraXiulan (approved):

src/Qowaiv/Text/RegOptions.cs
Corniel: Yes I do.

Issue #288, opened 2023-01-08
[Bug? or Documentation outdated]  Type conversion for int --> month not working


jeffward01:
Hello,

I am working on some basic things such as comparing the current month from `DateTime.UtcNow.Month` to `Month` in the `Qowaiv` library. 

I am unsure how to do this cleanly using the library, so I have referred to the docs.  In the docs I see under **Month**

```csharp
    [Fact]
    public void ShouldConvertToMonthInt()
    {
        Month feb = Month.Parse("February", CultureInfo.CurrentCulture);
        Month may = Month.May;
       
        // This below will not compile
        Month dec = 12;

        Assert.Equal(dec, Month.December);
    }
}
```

However, this does not compile due to a type converter not implemented.

Instead, I must do this:

```csharp
    [Fact]
    public void ShouldConvertToMonthInt()
    {
        Month feb = Month.Parse("February", CultureInfo.CurrentCulture);
        Month may = Month.May;
        Month dec = (Month)12;

        Assert.Equal(dec, Month.December);
    }
```

Questions:

1.)  Are the docs outdated?
2) Are there any recommended ways to compare `DateTime` with `Year` or `Month`

Thanks

---

![image](https://user-images.githubusercontent.com/12589359/211222839-b8350b0d-197e-45fb-8659-e57cd5bb030a.png)
[Link to docs](https://github.com/Qowaiv/Qowaiv#month)

Corniel:
Thanks for spotting an issue in the documentation. Those implicit castings haven been dropped (see #212) when introducing version 6.

To create a `Month` from an int you can use `Month.Create(int)` or indeed an explicit cast.

I've considered adding a `GetMonth()` extension on a `DateTime` and ``DateOnly` (and on the date types in Qowaiv). I still can do that. An other extesion could be to implement `IEquality<T>` on month (and year) for mutiple date/date time types, and extensions on `DateTime`. That can both be `Equals(Month month)` or, `HasMonth(Month month)`. From a domain-driven perspective you could argue that the second one has the better naming.

Can you tell me something about your use case?

jeffward01:
[...]

PR #290, opened 2023-01-11
Implement InYear and InMonth for date types

Corniel:
As requested by #288, it should be easy to check if a certain date (time) type is in a specific month or year.

I'm not sure yet of it should be `date.In(Year)` and `date.In(Month)` or `date.InYear(Year)` and `date.InYear(Month)`.


ahofland:
It depends a bit on what the practical application is, but wouldn't you also want to support ints in both functions? There is of course then an assumption (that the year is AC) but normally that's what you want.
Also, for the current implementation, .in(Year) and .in(Month) will be fine, as the type implies the usage. If you'd want to support ints, it'd need to be .InMonth(int) and .InYear(int) otherwise you don't have a way to separate them.

Corniel:
> It depends a bit on what the practical application is, but wouldn't you also want to support ints in both functions? There is of course then an assumption (that the year is AC) but normally that's what you want. Also, for the current implementation, .in(Year) and .in(Month) will be fine, as the type implies the usage. If you'd want to support ints, it'd need to be .InMonth(int) and .InYear(int) otherwise you don't have a way to separate them.

If you take the API of `System.DateTime` you see both `DateTime.Add(TimeSpan)` and `DateTime.AddMonths(int)`, so with that in mind you could argue that `.In(Year)` and `.In(Month)` is the best naming.

I'm not even sure if `.InYear(int)` should be supported: What does it add compared to `DateTime.Year == year`? For the `Year` and `Month` SVO on the otherside, There are downsides on supporting those overloads.

pmdevers:
Mayby we should not add it to the Svo it self but use extention methods. Because it does not add functionality to the svo.

Corniel:
> Mayby we should not add it to the Svo it self but use extention methods. Because it does not add functionality to the svo.

What do you want to gain with that?

src/Qowaiv/Extensions/System.DateOnly.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// <returns>
+    /// A new date whose value is the sum of the date and time represented
+    /// by this instance and the time interval represented by value.
+    /// </returns>
+    /// <exception cref="ArgumentOutOfRangeException">
+    /// The resulting date is less than <see cref="DateOnly.MinValue"/> or greater
+    /// than <see cref="DateOnly.MaxValue"/>.
+    /// </exception>
+    [Pure]
+    public static DateOnly Add(this DateOnly d, DateSpan value, bool daysFirst)
```
Sjaaky, line 51: I see how this method solves a problem. But I don't think it makes clear what happens to the reader of calling code.

Corniel: This is the same API as for Date (and DateTime). So changing is a breaking change.

src/Qowaiv/Date.cs
```diff
@@ -370,7 +368,19 @@ public Date Add(DateSpan value, bool daysFirst)
     [Pure]
     public Date AddMilliseconds(double value) => new(m_Value.AddMilliseconds(value));
 
-    #endregion
+    /// <summary>Returns true if the date is in the specified month, otherwise false.</summary>
+    /// <param name="month">
+    /// The <see cref="Qowaiv.Month"/> the date should be in.
+    /// </param>
+    [Pure]
+    public bool In(Month month) => !month.IsEmptyOrUnknown() && Month == (int)month;
```
Sjaaky, line 376: I prefer IsIn(...) to emphasize that it is a boolean check.

Corniel: That is a good suggestion!

Corniel:
> > > Mayby we should not add it to the Svo it self but use extention methods. Because it does not add functionality to the svo.
> > 
> > 
> > What do you want to gain with that?
> 
> Methods that do not change the state of the Svo can be placed outside. I believe this will improve readability of a class.

All SVO's are immutable. All methods are decorated as `[Pure]`, so then all methods would become extension methods. I don't think that would help anything here?!

Sjaaky:
> Methods that do not change the state of the Svo can be placed outside. I believe this will improve readability of a class.

The purpose of extension methods is to easily extend classes that are not under your control, to extend interfaces with common functionality. And could also be used to allow for a nullable 'this'.
Since the Svo is under our control, the best place is in the Svo itself. For readability we can use regions and partial classes.

Sjaaky (approved):
lgtm

src/Qowaiv/Extensions/System.DateOnly.cs
Corniel: This has been addressed with #290 .

Issue #247, opened 2022-06-04
Should IsValid(string) return false for values representing unknown


Corniel:
Currently, the static methods `IsValid(string str)` return false for `null`, white-space, and if parsing that input would fail. However, with the same argument that a value representing the empty state is not valid (in the context of this method), you would argue that the unknown state should neither be valid.

What would you think?

Neverfor:
I think that we might all agree, that state is not the same as value.

State can only be switched (assigned is a better word, but it is also used for values). In case, when it is not present, it is probably Unknown or Not Set, which is also a state.

When we talk about checking the input, then it becomes a value.

The check for IsValid (String.Empty) should therefore be based on:
> what is being checked: state or value (in our case, it's probably value)
> dotNet version
> if the setting used for nullables is enabled (with this setting, you see the warning, when it is possible for a nullable value to be passed to a non-nullable field).

Therefore:
if IsValid(string.Empty) and if IsValid(null) should return true for dotNet lower than 6 (I think it was 6, where they added it, but might have been 5).

if isValid(string.Empty) and if IsValid(null) should return false for dotNet 6 and above, unless Nullable setting is enabled for the project.

Why?
Becuase it will function the same way dotNet and Visual Studio work: it is possible to set a nullable value, but a warning will be given. It should also allow an intentional choice in architecture, where an empty prameter may be given and it must result in an exception.

Depending on the compiler, might be a bit crude, but the following compiles just fine:

string stringEmpty = string.Empty;
string stringNull = null;
string? stringAbsoluteNull = null;

Console.WriteLine(stringEmpty);
Console.WriteLine(stringNull);
Console.WriteLine(stringAbsoluteNull);

While the following will result in an error:

string nothing;
Console.WriteLine(nothing);

Yet, this will compile, if we explicitly set the object value, or rather - change the object state, because the value will remain the same. From the user's point of view, the value is set to another value, which happens to be the same (null), though: 

string nothing;
nothing = null;
Console.WriteLine(nothing);

If we look at this specific case, we can say that important part is the state. Ideally, you would also check if the item was updated, after it's creation. But this would go beyond the simple value check.

pmdevers:
I think there is no right answer. It all depends on the buisness requirements what is valid. Therefor I would suggest to entirely drop the IsValid methods because of this.

Corniel:
[...]

PR #297, opened 2023-01-17
Mark all static IsValid(string) methods obsolete

Corniel:
The result of the #247 issue is that is better to get rid of the `.IsValid(string)` and .IsValid(string, CulutreInfo) static methods. If Empty and/or Unknown should be included, is a domain decision, that outside the scope of these methods. Furthermore, the actual usage of the SVO's should not be discouraged.

### DIFF
--- a/src/Qowaiv.Data.SqlClient/Generated/Timestamp.generated.cs
+++ b/src/Qowaiv.Data.SqlClient/Generated/Timestamp.generated.cs
@@ -239,6 +239,7 @@ public partial struct Timestamp
     /// The <see cref="string"/> to validate.
     /// </param>
     [Pure]
+    [ExcludeFromCodeCoverage]
     [Obsolete("Use the SVO itself instead. Will be dropped when the next major version is released.")]
     public static bool IsValid(string? val) => IsValid(val, (IFormatProvider?)null);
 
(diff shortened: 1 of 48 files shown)


src/Qowaiv.Data.SqlClient/Generated/Timestamp.generated.cs
```diff
@@ -239,6 +239,8 @@ public partial struct Timestamp
     /// The <see cref="string"/> to validate.
     /// </param>
     [Pure]
+    [ExcludeFromCodeCoverage]
+    [Obsolete("Use the SVO itself instead. Will be dropped when the next major version is released.")]
```
Sjaaky, line 243: Could you elaborate on 'Use the SVO itself'? I would love a clear piece of code, here, or in a readme file, with which I can replace my IsValid(str) calls.

Sjaaky (changes requested):

src/Qowaiv.Data.SqlClient/Generated/Timestamp.generated.cs
Corniel: Fair point. I Updated it to : `"Use Timestamp.TryParse(str) is { } instead"`.

Sjaaky (approved):

PR #300, opened 2023-01-22
Decorate methods that always throw with [DoesNotReturn] attribute

Corniel:
The `[DoesNotReturn]` attribute  should be used to decorate methods that never return (but always throw). 

## See
* https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.doesnotreturnattribute.
* https://endjin.com/blog/2020/08/dotnet-csharp-8-nullable-references-when-methods-dont-return


LauraXiulan (changes requested):
Converters should be sealed. Build generates warnings on the DoesNotReturn attribute for overridden members.

PR #301, opened 2023-02-04
Decorate string arguments representing regexes with [StringSyntax(StringSyntaxAttribute.Regex)]

Corniel:
By decorating arguments with `[StringSyntax(StringSyntaxAttribute.Regex)]`, the IDE will give visual hints on the regular expressions. This is available since .NET 7.0.

### DIFF
--- a/src/Qowaiv/OpenApi/OpenApiDataType.cs
+++ b/src/Qowaiv/OpenApi/OpenApiDataType.cs
@@ -57,6 +57,9 @@ public sealed record OpenApiDataType
     public bool Nullable { get; }
 
     /// <summary>Gets the Pattern of the OpenAPI Data Type.</summary>
+#if NET7_0_OR_GREATER
+    [StringSyntax(StringSyntaxAttribute.Regex)]
+#endif
     public string? Pattern { get; }
 
     /// <summary>Gets the Pattern of the OpenAPI Data Type.</summary>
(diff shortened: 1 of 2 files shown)


src/Qowaiv/OpenApi/OpenApiDataType.cs
```diff
@@ -53,6 +57,9 @@ public sealed record OpenApiDataType
     public bool Nullable { get; }
 
     /// <summary>Gets the Pattern of the OpenAPI Data Type.</summary>
+#if NET7_0_OR_GREATER
```
Sjaaky, line 60: Inconsistencies between 
#if net70 [attr] prop #else prop #endif
and
#if net70 [attr] #endif prop

I prefer the latter, maybe even on a single line.

Sjaaky: Another consideration could be to define a "dummy" StringSyntaxAttribute for dotnet 6 and lower. But for these 4 occurrences I wouldn't bother.

Sjaaky (approved):

src/Qowaiv/OpenApi/OpenApiDataType.cs
Corniel: @Sjaaky I considered the second option. But is both an attribute and a helper static class.

About `#if .+` The compiler does not support any code after the `#` so your suggestions would not compile.

Issue #303, opened 2023-02-17
Support async and await for Clock.SetThreadUtcNow()


Sjaaky:
Currently `Clock.SetTimeForCurrentThread` stores the `Func<DateTime>` in a `[ThreadStatic]` variable. This works perfectly for synchronous tests. 
```cs
using(Clock.SetTimeForCurrentThread(() => new DateTime(2023, 2, 17))
{
    // test code
}
```
But for tests that use async/await, code is executed on different threads. On those threads the Clock is not set.
To solve this, the `threadUtcNow` and `threadTimeZone` should be stored using `AsyncLocal<>`.

Question is, should the name of `SetTimeForCurrentThread` be changed? I can't think of a better name than the current name, it applies to more situations though.

Corniel:
`ForCurrentContext` is a good new suffix. However, that should then obviously only be a new way of doing it, and the old methods should be marked as depreciated.

Corniel:
Solved with #304

PR #304, opened 2023-02-17
Clock: Store time and timezone in AsyncLocal<> instead of ThreadStatic.

Sjaaky:
Make Clock also work with async/await.
Solves #303


Corniel:
@Sjaaky What happens now if multiple tests run in parallel with different using blocks?

Corniel (approved):
LGTM.

src/Qowaiv/Qowaiv.csproj
```diff
@@ -5,8 +5,14 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net5.0;net6.0;net7.0</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <Version>6.4.0</Version>
+    <Version>6.4.1</Version>
     <PackageReleaseNotes>
+v6.4.1
+- Clock.SetTimeForCurrentContext supports async. #304
+- Fix issue with not closed comment in parsing email adresses. #298
```
LauraXiulan, line 12: Are all these changes in this PR, or are they previous changes that didn't get a version bump yet?

Corniel: Yep. They where considered too minor to get a separate release.

src/Qowaiv/Clock.cs
```diff
@@ -36,14 +38,14 @@ public static class Clock
...
-        var utcNow = (threadUtcNow ?? globalUtcNow).Invoke();
+        var utcNow = (threadUtcNow.Value ?? globalUtcNow).Invoke();
         return utcNow.Kind == DateTimeKind.Utc
             ? utcNow
             : new DateTime(utcNow.Ticks, DateTimeKind.Utc);
     }
 
     /// <summary>Gets the time zone of the <see cref="Clock"/>.</summary>
-    public static TimeZoneInfo TimeZone => threadTimeZone ?? globalTimeZone;
+    public static TimeZoneInfo TimeZone => threadTimeZone.Value ?? globalTimeZone;
```
LauraXiulan, line 48: Can threadTimeZone be null?

Corniel: Yep. It can. I previously was a `Func` that could be null. Now it is a `LocalAsync<T>` where the `Value` can be null.

LauraXiulan (approved):

PR #305, opened 2023-02-20
Renaming of internals of Clock

Corniel:
The internals that previously set the thread (now local/context) related values have been renamed to match their current behaviour.

### DIFF
--- a/src/Qowaiv/Clock.cs
+++ b/src/Qowaiv/Clock.cs
@@ -38,14 +38,14 @@ public static class Clock
     [Pure]
     public static DateTime UtcNow()
     {
-        var utcNow = (threadUtcNow.Value ?? globalUtcNow).Invoke();
+        var utcNow = (localUtcNow.Value ?? globalUtcNow).Invoke();
         return utcNow.Kind == DateTimeKind.Utc
             ? utcNow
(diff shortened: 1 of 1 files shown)


LauraXiulan (approved):

src/Qowaiv/Clock.cs
```diff
@@ -38,14 +38,14 @@ public static class Clock
     [Pure]
     public static DateTime UtcNow()
     {
-        var utcNow = (threadUtcNow.Value ?? globalUtcNow).Invoke();
+        var utcNow = (localUtcNow.Value ?? globalUtcNow).Invoke();
```
Sjaaky, line 41: 'local' probably means something like local to the current "execution" context. But the term local is also used within the time domain wrt timezones and thus not very descriptive.
I'm struggling with clear terminologie as well. Some suggestions:
Alternative 1) Use the term `asyncLocalUtcNow`, to indicate it is local to the current execution context.
Alternative 2) localContextUtcNow,
Alternative 3) currentContextUtcNow

Corniel: @Sjaaky localContextUtcNow might be a better suggestion indeed.

PR #308, opened 2023-03-08
Use Spans to concatenate strings

Corniel:
As suggested by CA1845, use `.AsSpan()` for string manipulations.


Corniel:
@Sjaaky Wow, that is even a better way of solving the issue. LGTM.

PR #310, opened 2023-05-08
Use BackwardsCompatibleFeatures package to simplify build

Corniel:
As BackwardsCompatibleFeatures reduces the need to some `#if` code blocks. More details can be found [here](https://github.com/CptWesley/BackwardsCompatibleFeatures).


CptWesley (approved):
LGTM

PR #317, opened 2023-07-30
Drop not used dependencies

Corniel:


Corniel:
Abandoned, we switched to PolySharp.

Issue #321, opened 2023-07-31
Qowaiv 4.6.2 depends on DotNetProjectFile.Analyzers


Sjaaky:
Some of our unittests failed with this error:
`System.TypeLoadException : Declaration referenced in a method implementation cannot be a final method.  Type: 'Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions'.  Assembly: 'Microsoft.CodeAnalysis.CSharp, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.`
Which indicates a mismatch between the currently loaded type and the type against which a library was compiled.

The build also gave a warning like this:
`CS9057	The analyzer assembly '...\packages\dotnetprojectfile.analyzers\1.0.7\analyzers\DotNetProjectFile.Analyzers.dll' references version '4.6.0.0' of the compiler, which is newer than the currently running version '4.5.0.0'.`

After [some digging](https://www.nuget.org/packages/Qowaiv/6.4.2#dependencies-body-tab) Qowaiv 4.6.2 seems to reference dotnetprojectfile.analyzers. Since dotnetprojectfile.analyzers uses a newer .net version than our project (we use net6.0), this leads to runtime errors.

I don't think that Qowaiv should depend on the DotNetProjectFile.Analyzers.

Corniel:
Shipped: https://www.nuget.org/packages/Qowaiv/6.4.3

PR #322, opened 2023-07-31
Analyzers should not be included in NuGet packages

Corniel:
By cleaning-up the props, two analyzers where not longer marked as `PrivateAssets`.

See: #321


Sjaaky (approved):
lgtm

PR #324, opened 2023-08-07
Introduction fo the (EU) Energy label

Corniel:
Specially due to the `+`-es, this is preferable over an enum.


src/Qowaiv/Conversion/Sustainability/EnergyLabelTypeConverter.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Qowaiv.Sustainability;
+
+namespace Qowaiv.Conversion.Sustainability;
+
+/// <summary>Provides a conversion for a Sex.</summary>
```
LauraXiulan, line 5: Wrong summary

specs/Qowaiv.Specs/Sustainability/EnergyLabel_specs.cs
```diff
@@ -0,0 +1,521 @@
...
+        var formatted = Svo.EnergyLabel.ToString("SomeFormat", FormatProvider.CustomFormatter);
+        Assert.AreEqual("Unit Test Formatter, value: 'A++', format: 'SomeFormat'", formatted);
+    }
+
+    [TestCase(null, "A++", "A++")]
+    [TestCase("U", "B", "B")]
+    [TestCase("l", "G", "g")]
+    [TestCase("l", "A++", "a++")]
+    [TestCase("L", "A++", "a++")]
+    public void fomrat_dependent(string format, EnergyLabel svo, string formatted)
```
LauraXiulan, line 275: format

specs/Qowaiv.Specs/Sustainability/EnergyLabel_specs.cs
```diff
@@ -0,0 +1,521 @@
...
+    public void can_be_sorted_using_compare()
+    {
+        var sorted = new[]
+        {
+            default,
+            default,
+            EnergyLabel.F,
+            EnergyLabel.C,
+            EnergyLabel.A(),
+            EnergyLabel.Unknown,
```
LauraXiulan, line 305: Unknown is always highest? I would have expected it after the default

Corniel: default/Empty comes direct before G, so that would be strange.

PR #327, opened 2023-08-24
Introduction of `HasValue` and `IsKnown` properties

Corniel:
When Qowaiv started, it was a considered decision to make `.IsEmpty()` ,`.IsUnknown()`, and `.IsEmptyOrUnknown()` methods, rather then properties of the SVO's. After all, they are only partly related to those SVO domains.

That being said, with the introduction of [#C pattern matching](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching) it would be nice to check for those states via properties, in one way, or another.

To allow pattern matching, and not to change those long lived methods, this is the proposal:

- `.HasValue` as the functional equivalent of `!IsEmpty()`
- `.IsKnown` as the functional equivalent of `!IsEmptyOrUnknown()` 

This can be used like this:

``` C#
if (EmailAddress.TryParse(str) is { IsKnown: true } email(
{
    ...
}
```
 
 and obviously other pattern matching scenario's.

### DIFF
--- a/specs/Qowaiv.Specs/Chemistry/CAS_Registry_Number_specs.cs
+++ b/specs/Qowaiv.Specs/Chemistry/CAS_Registry_Number_specs.cs
@@ -4,12 +4,21 @@ namespace Chemistry.CAS_Registry_Number_specs;
 
 public class With_domain_logic
 {
+    [TestCase(true, "73–24–5")]
+    [TestCase(true, "?")]
+    [TestCase(false, "")]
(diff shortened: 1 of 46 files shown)


LauraXiulan:
Just a consideration, would HasValue lead to confusion compared to the Microsoft Nullable HasValue property? As that one can still have an 'empty' value, but this one is specifically not an empty value.

pmdevers (approved):

PR #336, opened 2023-11-08
Extend FormatException with detailed information for failed parsings

Corniel:
When parsing fails, it might be useful to have detailed information on which string value could not be parsed to which type. The latter is particularly useful when applied to generics such as `Id<T>` .

To prevent different exceptions being thrown, and messages changes, this is solved with adding an inner exception that contains this extra information. By doing so, the `ToString()` of the exception is extended with this information, but the default message and type are kept the same.


Sjaaky (approved):
lgtm

Issue #338, opened 2023-11-16
Explicit support for DateOnly on MonthSpan and DateSpan


ROSDEV:
While using MonthSpan or DateSpan for example. 

MonthSpan.Subtract((Date)dateOnlyDate, (Date)dateOnlyDate)

I need to explicit cast the params as Qowaiv.Date. Can there be support / overloads for Dateonly type. 

Preferable for all factory methods.

PR #339, opened 2023-11-17
All public methods with an `Date` parameter should have a `DateOnly` overload

Corniel:
Since .NET 6.0, the use of `DateOnly` is preferred over the use of `Date` (a type that was created because .NET did not have an alternative back in the day). To improve the usability of the API methods requiring an `Date` should have an overload with a `DateOnly` too. #338


specs/Qowaiv.Specs/Date_span_specs.cs
```diff
@@ -107,3 +108,145 @@ public void with_info()
...
+        plussed.Should().Be(Svo.DateSpan);
+    }
+}
+
+public class Can_create
+{
+    [Test]
+    public void Age_form_date_without_months()
+    {
+        using (Clock.SetTimeForCurrentContext(() => new Date(2019, 10, 10)))
```
ROSDEV, line 134: Suggest use [UseTime] Attribute

ROSDEV (approved):

specs/Qowaiv.Specs/Date_span_specs.cs
Corniel: That option does not exist in this repo. ;)

PR #341, opened 2023-11-21
ICustomFormatProvider should only be applied when it does not return null

Corniel:
An `ICustomFormatProvider` should  not be enforced to know how to format all types. It can indicate this be return `null`.


renzo-baasdam (approved):

Sjaaky:
lgtm

PR #349, opened 2023-12-21
Extend IBAN support

Corniel:
Added specific support for Russia, Sudan, and the Central African Republic. Because Russia's IBAN has a length of 33, it was rejected by the generic validation assuming that the maximum length was 32. This has been changed to 36.

Burundi's pattern turned out to be wrong.


IBAN.md
```diff
@@ -0,0 +1,126 @@
+﻿# International Bank Account Number
+An overview of all IBAN's supported. Note that for countries not mentioned, IBA
```
LauraXiulan, line 2: Missing text? The sentence seems to stop abruptly

IBAN.md
```diff
@@ -0,0 +1,126 @@
...
+ISO 7064 MOD-97-10
+
+# Basic Bank Account Number
+Abbreviated as BBAN, is a way of describing the IBAN (for a specific country).
+* a: alphabetic
+* c: alphanumeric 
+* n: numeric
+
+In this table, for countries with fixed characters (such as Costa Rica) those
+are displayed as `[?]`. This is strictly not BBAN, but helps when genrating
```
LauraXiulan, line 13: generating

src/Qowaiv/Financial/InternationalBankAccountNumber.Patterns.cs
```diff
@@ -19,16 +19,17 @@ namespace Qowaiv.Financial;
         Bban(Country.AL, "8n,16c"),
         Bban(Country.AO, "21n"),
         Bban(Country.AT, "16n"),
-        Bban(Country.AZ, "4c,20n"),
+        Bban(Country.AZ, "4a,20c"),
```
LauraXiulan, line 22: Why so many changes in patterns, the description of this PR only mentions four countries?

LauraXiulan (changes requested):
See other comments

src/Qowaiv/Financial/InternationalBankAccountNumber.Patterns.cs
Corniel: I did not want to mention this kind of (relatively) small changes. From numeric to alphanumeric obviously ment that previously IBAN's could have been rejected incorrectly (the other way around is not an issue). But you've got a point, these are also changes.

IBAN.md
Corniel: Is this better?

LauraXiulan: Yes

LauraXiulan (approved):

PR #353, opened 2024-01-04
Allow IBAN's to be defined with an IBAN/IBAN:/(IBAN) prefix

Corniel:
Parsing is about being forgiving on the input. It turns out, that both `"(IBAN) NL20INGB0001234567"`, `"IBAN NL20INGB0001234567"`, and ``"IBAN: NL20INGB0001234567"`` are not uncommon ways of representing IBAN's. Therefor, with this PR, those prefixes are allowed (and removed from) input strings.

### DIFF
--- a/src/Qowaiv/Financial/IbanParser.cs
+++ b/src/Qowaiv/Financial/IbanParser.cs
@@ -14,6 +14,7 @@ internal static partial class IbanParser
         var index = 0;
         var pos = 0;
         var id = 0;
+        var prefixed = false;
 
         while (index < str.Length)
         {
@@ -30,11 +31,12 @@ internal static partial class IbanParser
                 {
                     return bban.Parse(str, index, id);
                 }
-                else if (HasIbanPrefix(id, str, index))
+                else if (!prefixed && HasIbanPrefix(id, str, index))
                 {
                     pos = 0;
                     id = 0;
                     index += 3;
+                    prefixed = true;
                 }
                 else return null;
             }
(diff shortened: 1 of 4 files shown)


src/Qowaiv/Financial/IbanParser.cs
```diff
@@ -12,24 +14,42 @@ internal static partial class IbanParser
         var index = 0;
         var pos = 0;
         var id = 0;
+        var prefixed = false;
 
-        foreach (var ch in str)
+        while (index < str.Length)
```
renzo-baasdam, line 19: The while loop is a bit complicated. Am wondering is the following idea is possible:
- Loop till no markup
- Check if there is one IBAN prefix (either (IBAN) or IBAN, first 2 chars would be enough to call function) -> move to after prefix
- Loop till no markup
- Use Bban to parse rest of string

Not sure if this would prevent a double prefix.

Corniel: Your approach would slow down all parsing, while prefixes are relatively uncommon. With this approach,  only if the default flow does not result in a valid IBAN (a valid country code to start with), a prefix check kicks in. It also does not require to check the string multiple times or creating substrings.

renzo-baasdam: I do think the above would only require one passthrough, but perhaps it would not optimize default case.
Also I think that `str[(index - 2)..]` and `str[(index - 1)..]` do create copies in the current implementation.

PR #360, opened 2024-01-16
Align Parse and TryParse provider naming with IParsable

Corniel:
As pointed out by @pmdevers  in #281 (and #273): The `IParsable<TSelf>` contract specifies the `IFormatProvider` arguments as `provider`, not `formatProvider`. As this is a beaking change it has been postponed until now.


Corniel:
> Should we not add IParsable Interface if net8 or higher

That is already in place, hence the warnings.

pmdevers (approved):

PR #365, opened 2024-01-18
Improve performance of UUID parsing

Corniel:
Parsing UUID's (the base64 string representation) was dramatically slower than parsing a regular `GUID`. By dropping the `Regex`, using an `unsafe` cast (trick that MS does as well), times improved.

By reducing the function calls (for all parse methods) durations where even reduced more (basically 12 ns).

After these changes, the overhead for using `UUID` instead of MS's `GUID` is (almost) neglectable.

| Parse              | Mean     | Ratio |
|------------------- |---------:|------:|
| GUID (Guid.Parse)  | 15.63 ns |  1.00 |
| GUID               | 18.77 ns |  1.20 |
| Base64             | 18.58 ns |  1.19 |
| Base32             | 27.82 ns |  1.78 |
| Regex + FromBase64 | 92.76 ns |  5.94 |

ToString has also been improved, but is still 57% slower than writing a GUID.

| ToString             | Mean     | Ratio |
|--------------------- |---------:|------:|
| GUID (Guid.ToString) | 11.10 ns |  1.00 |
| Base64               | 17.47 ns |  1.57 |
| GUID                 | 43.03 ns |  3.85 |
| Convert.ToBase64     | 50.71 ns |  4.57 |

The version of a `UUID` is stored in the upper 4 bits of byte 7. By using the
`GuidLayout` of `ToByteArray()` to retrieve that those bits is a big improvement as well.

| Method      | Mean       | Ratio |
|------------ |-----------:|------:|
| Layout      |   0.584 ns |  1.00 |
| From byte[] |   3.024 ns |  5.20 |

### DIFF
--- a/specs/Qowaiv.Benchmarks/Program.cs
+++ b/specs/Qowaiv.Benchmarks/Program.cs
@@ -6,9 +6,10 @@ public static class Program
 {
     public static void Main()
     {
+        BenchmarkRunner.Run<UuidBenchmark.Parse>();
         BenchmarkRunner.Run<UuidBenchmark.StringOutput>();
-        //BenchmarkRunner.Run<UuidBenchmark.Version>();
-        //BenchmarkRunner.Run<IbanBenchmark.ParseUnformatted>();
-        //BenchmarkRunner.Run<IbanBenchmark.ParseFormatted>();
+        BenchmarkRunner.Run<UuidBenchmark.Version>();
+        BenchmarkRunner.Run<IbanBenchmark.ParseUnformatted>();
+        BenchmarkRunner.Run<IbanBenchmark.ParseFormatted>();
     }
 }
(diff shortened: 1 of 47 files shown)


specs/Qowaiv.Benchmarks/Program.cs
```diff
@@ -6,7 +6,9 @@ public static class Program
 {
     public static void Main()
     {
-        BenchmarkRunner.Run<IbanBenchmark.ParseUnformatted>();
-        BenchmarkRunner.Run<IbanBenchmark.ParseFormatted>();
+        BenchmarkRunner.Run<UuidBenchmark.StringOutput>();
+        //BenchmarkRunner.Run<UuidBenchmark.Version>();
```
CptWesley, line 10: Are these commented out intentionally?

CptWesley (approved):

PR #366, opened 2024-01-19
Introduction of IUnknown<TSelf>

Corniel:
A (partly static) contract to indicates that the SVO's default value represents an unknown (but set) state.

### DIFF
--- a/src/Qowaiv/Hashing/HashingNotSupported.cs
+++ b/src/Qowaiv/Hashing/HashingNotSupported.cs
@@ -5,9 +5,7 @@
 /// </summary>
 [ExcludeFromCodeCoverage]
 [Serializable]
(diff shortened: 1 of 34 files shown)


src/Qowaiv/Hashing/HashingNotSupported.cs
```diff
@@ -5,7 +5,9 @@
 /// </summary>
 [ExcludeFromCodeCoverage]
 [Serializable]
+#pragma warning disable S3925 // "ISerializable" should be implemented correctly
 public class HashingNotSupported : NotSupportedException
+#pragma warning restore S3925 // "ISerializable" should be implemented correctly
```
LauraXiulan, line 10: Can be removed, warning is put on none

LauraXiulan (approved):

PR #367, opened 2024-01-20
Implement INumber on Percentage

Corniel:
By implementing `INumber<Percentage>` the usage of `Qowaiv.Percentage` can be even further improved. I've chosen to implement most new members explicit, but am open for suggestions on which to implement implicit.


CptWesley (approved):
LGTM

PR #372, opened 2024-01-29
Implement IParsable for Id<TIdBehavior>

Corniel:
It turned out that `Id<T>` did not implement `IParsable` was missing. Easy fix.

### DIFF
--- a/src/Qowaiv/Identifiers/Id.cs
+++ b/src/Qowaiv/Identifiers/Id.cs
@@ -21,6 +21,7 @@ namespace Qowaiv.Identifiers;
 public readonly struct Id<TIdentifier> : IXmlSerializable, IFormattable, IEquatable<Id<TIdentifier>>, IComparable, IComparable<Id<TIdentifier>>
 #if NET7_0_OR_GREATER
 , IEqualityOperators<Id<TIdentifier>, Id<TIdentifier>, bool>
+, IParsable<Id<TIdentifier>>
 #endif
 #if NET8_0_OR_GREATER
 #else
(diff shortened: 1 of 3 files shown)


src/Qowaiv/Identifiers/Id.cs
```diff
@@ -21,6 +21,7 @@ namespace Qowaiv.Identifiers;
 public readonly struct Id<TIdentifier> : IXmlSerializable, IFormattable, IEquatable<Id<TIdentifier>>, IComparable, IComparable<Id<TIdentifier>>
 #if NET7_0_OR_GREATER
 , IEqualityOperators<Id<TIdentifier>, Id<TIdentifier>, bool>
+, IParsable<Id<TIdentifier>>
```
LauraXiulan, line 24: IParsable is added to net7 or greater, but the test is specifically for net8 or greater, is that correct?

src/Qowaiv/Identifiers/Id.cs
```diff
@@ -240,6 +241,24 @@ void IXmlSerializable.WriteXml(XmlWriter writer)
...
+    /// </param>
+    /// <returns>
+    /// <returns>
+    /// The parsed identifier.
+    /// </returns>
+    /// <exception cref="FormatException">
+    /// <paramref name="s"/> is not in the correct format.
+    /// </exception>
+    [Pure]
+    public static Id<TIdentifier> Parse(string s, IFormatProvider? provider)
```
LauraXiulan, line 259: If IParsable is only for net7 or greater, should these methods also be inside an if?

src/Qowaiv/Identifiers/Id.cs
Corniel: Yes. This is a patch as @CptWesley asked for this with urgency. But as Qowaiv 7 is in the making, I just would add more work to apply those changes later on.

src/Qowaiv/Identifiers/Id.cs
Corniel: For the other SVO's (where this is generated) those overloads are also available for all targets.

LauraXiulan (approved):

PR #374, opened 2024-02-13
 Deserializing ID's from JSON numeric nodes takes constrains into account

Corniel:
When an (numeric) identifier was deserializing, the constraints defined (in `TryCreate`) where not triggered.

### DIFF
--- a/specs/Qowaiv.Specs/Identifiers/Id_for_Int32_specs.cs
+++ b/specs/Qowaiv.Specs/Identifiers/Id_for_Int32_specs.cs
@@ -49,6 +49,37 @@ public class Supports_JSON_serialization
     [TestCase("12345678", 12345678L)]
     public void System_Text_JSON_serialization(Int32Id svo, object json)
         => JsonTester.Write_System_Text_JSON(svo).Should().Be(json);
+
+    [TestCase(-2)]
+    [TestCase(17)]
+    [TestCase("17")]
+    [TestCase(int.MaxValue + 1L)]
+    public void taking_constrains_into_account(object json)
+    {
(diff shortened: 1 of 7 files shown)


specs/Qowaiv.Specs/Identifiers/Id_for_Int32_specs.cs
```diff
@@ -49,6 +49,37 @@ public void System_Text_JSON_deserialization(object json, Int32Id svo)
...
+    private sealed class ForEven : Int32IdBehavior
+    {
+        public override bool TryCreate(object? obj, out object? id)
+        {
+            if(obj is int even && even % 2 == 0)
+            {
+                id = even;
+                return true;
+            }
+            {
```
renzo-baasdam, line 73: Does this work without else?

specs/Qowaiv.Specs/Identifiers/Id_for_Int64_specs.cs
```diff
@@ -48,7 +48,37 @@ public void System_Text_JSON_deserialization(object json, Int64Id svo)
...
+    private sealed class ForEven : Int64IdBehavior
+    {
+        public override bool TryCreate(object? obj, out object? id)
+        {
+            if (obj is long even && even % 2 == 0)
+            {
+                id = even;
+                return true;
+            }
+            {
```
renzo-baasdam, line 71: Idem

specs/Qowaiv.Specs/Identifiers/Id_for_Int32_specs.cs
Corniel: It was unintended, but yes, it works. `{ .. }` just defines a scope.

renzo-baasdam (approved):

Issue #379, opened 2024-03-12
Revisit build process


Corniel:
Things currently going wrong:

* Sonar complains about the Java version
* Build seems to fail due to not be able to load a dll
* Build claims to succeeded while all tests fail
* Insight in tests is lacking (for inspiration see [Buildalyzer](https://github.com/phmonte/Buildalyzer/tree/main/.github/workflows))

PR #381, opened 2024-03-18
Improve pipeline

renzo-baasdam:
Working on #379.
- [x] _Convert container to Linux_
- [x] _Sonar complains about the Java version_
Cannot recreate warning.
- [x] _Build seems to fail due to not be able to load a dll_
Cannot recreate warning.
- [x] _Build claims to succeeded while all tests fail_
Fixed by separating a multi line command into 3 separate steps.
- [X] _Insight in tests is lacking (for inspiration see [Buildalyzer](https://github.com/phmonte/Buildalyzer/tree/main/.github/workflows))_
Seems to kindof work despite exceeding allowed size
`  Test report summary is too big - setting 'listTests' to 'failed'
`


Corniel:
I added a task: convert (back) to Linux. This might include running the *netstandard2.0* unit test targets as .NET 4.8 instead over 3.1 core.

.github/workflows/code-coverage.yml
```diff
@@ -1,9 +1,5 @@
 name: Code coverage
 on:
-  push:
-    branches:
```
Corniel, line 4: It should be on master though.

renzo-baasdam: Will revert, was testing some things.

Corniel: Sure. Just as a reminder. Speaking of which, That show test results is a new build step/script. But I did not see the results, is that still work in progress, or did checked the wrong place?

.github/workflows/build-test.yaml
```diff
@@ -3,66 +3,73 @@ on: [push]
...
+        if: success() || failure()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        run: |
+          ./.sonar/scanner/dotnet-sonarscanner end \
+          /d:sonar.token="${{ secrets.SONAR_TOKEN }}"
+      - name: Upload test results
+        uses: actions/upload-artifact@v4
+        if: success() || failure()
```
Corniel, line 72: What is the reasoning about this? this seems always to be true, but I guess it is not that trivial?

.github/workflows/code-coverage.yml
renzo-baasdam: It is indeed still in progress. The problem with the method I am trying (waiting for the `Build` workflow to complete and then using the results in a new workflow) is that it only works when that workflow is on `master` (see https://github.com/orgs/community/discussions/25288). And since it isn't there, it doesn't show currently.

Corniel:
LGTM! Thanks for effort.

PR #382, opened 2024-03-27
Drop EmailAddressCollection

Corniel:
EmailAddressCollection is not a pure SVO and its behaviour does not add enough value to keep maintaining it.


src/Qowaiv/Qowaiv.csproj
```diff
@@ -9,6 +9,7 @@
     <PackageId>Qowaiv</PackageId>
     <PackageReleaseNotes>
 v7.0.0
+- Drop email address collection. EmailAddression will be removed: #382 (breaking)
```
renzo-baasdam, line 12: EmailAddressCollection

renzo-baasdam (approved):

PR #386, opened 2024-04-19
JSON-deserialization from decimal.MaxValue (as double) should not fail

Corniel:
When `Amount.Max` is serialized to JSON, its gets the value `7.922816251426434E+28`. This value can not be deserialized, because of rounding/precision issues, that can be demonstrated without any JSON serialization:

``` C#
var dbl = (double)decimal.MaxValue;
var dec = (decimal)dbl;
```

This code fails on the second line saying that `dbl` is either too big or too small.

This p.r. adds some forgiveness when casting back from doubles to decimals, that prevents this issue from occurring.


CptWesley (approved):

renzo-baasdam (approved):

Sjaaky (approved):
lgtm

PR #394, opened 2024-07-03
Reduce trailing zero's by Percentage manipulations 

Corniel:
With the introduction of Qowaiv  v7, Some `Percentage` manipulations started to create more trailing zero's than they did before. In most cases this will not be noticed, but at least one scenario is affected:

``` C#
var p = somePercentage * Percentage.Hundred; // this obviously could be the outcome of some decision tree.
JsonSerializer.Seralize(p);
```

The result of the serialization has 2 more trailing zero's than it had with Qowaiv v6.6.2 (and earlier versions). With the introduction of a (for now internal) `DecimalMath` that can change the Scale (moving the decimal separator back- or forward) and apply a `Trim`, these issues are dramatically reduced.

While at it, I noticed that the allowed boundaries of `Percentage` (those of `decimal` divided by 10.000) where not guarded. Now they are.

### DIFF
--- a/specs/Qowaiv.Benchmarks/DecimalBenchmark.cs
+++ b/specs/Qowaiv.Benchmarks/DecimalBenchmark.cs
@@ -1,0 +1,26 @@
+﻿using Qowaiv;
+
+namespace Benchmarks;
+
+public class DecimalBenchmark
+{
+    public class Scale
+    {
+        [ParamsSource(nameof(Values))]
+        public decimal Value { get; set; }
+
+        public static decimal[] Values => [100, 3.14m, 23.4326m];
+
+        [Benchmark]
+        public Percentage DecimalMath()
+            => Value.Percent();
+
+        [Benchmark]
+        public Percentage Division()
+            => Percentage.Create(Value / 100m);
+
+        [Benchmark]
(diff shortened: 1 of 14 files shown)


CptWesley:
Wouldn't it have been easier and less error prone to simply replace to `.ToString()` implementation with a call to `.ToString("0.############################")` (28 `#`s to deal with the maximum scale of 28) on the underlying decimal? Also has the nice side effect of preventing unintentional scientific notation.

CptWesley:
To add to that:

```csharp
var dec = 0.000001000d;
Console.WriteLine(dec);  // 1E-06
Console.WriteLine(dec.ToString("0.############################")); // 0,000001 (with nl-NL culture)
Console.WriteLine(dec.ToString("0.############################")); // 0.000001 (with en-US culture)
```

Corniel:
@CptWesley I've applied a rebase. Most of the (previously) 'added' code, already existed to support custom decimal rounding. Now it should be easier to do the review.

About your example, for serializing `Percentage`s, your method is already applied. In the described scenario, is is about serializing decimal based types (`Amount` , and `decimal` itself) after being multiplied with a percentage.

PR #398, opened 2024-07-15
Enable package validation

Corniel:
See [Proj0240](https://dotnet-project-file-analyzers.github.io/rules/Proj0240.html).


Corniel:
Not yet.

Issue #407, opened 2024-08-12
Percentage round trip not possible


CptWesley:
Currently doing a round-trip of serializing and deserializing `Percentage.MaxValue` results in an error indicating that the value is too large to store in a `Percentage`.

PR #408, opened 2024-08-12
Fix percentage boundaries checks on parsing

Corniel:
The re-scale of parsed decimal was not taken into account *before* checking the boundaries. See #407.


krakkepit:
Looks like the fix you need!

PR #412, opened 2024-09-13
Use coverlet.collector

### LINKED ISSUE #411: Improve Coverlet usage
According to their [own documentation](https://github.com/coverlet-coverage/coverlet): 
> Do not add `coverlet.collector` and `coverlet.msbuild` package in a test project.

Which is precisely what we currently do. They also recommend to use `coverlet.collector`, but I guess, that that also implies that we should change the way we [generate our coverage files](https://github.com/Qowaiv/Qowaiv/blob/master/.github/workflows/code-coverage.yml) that we send to [coveralls.io](https://coveralls.io/github/Qowaiv/Qowaiv?branch=master)

renzo-baasdam:
Some pipeline improvements:
- Use `coverlet.collector` and remove `coverlet.msbuild` as the former is recommended there should not be two packages. Resolves #411.
- Fix readme build icon.

Note: the option `--logger "trx;LogFileName=test-results.trx"` together with `--collect:"XPlat Code Coverage;Format=opencover"` causes duplicate `coverage.opencover.xml` files. This is intended behavior apparently, upon checking SonarCloud it seems to work.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -50,15 +50,13 @@ jobs:
           /o:"qowaiv" \
           /d:sonar.token="${{ secrets.SONAR_TOKEN }}" \
           /d:sonar.host.url="https://sonarcloud.io" \
-          /d:sonar.cs.opencover.reportsPaths="**/coverage.**.opencover.xml" \
+          /d:sonar.cs.opencover.reportsPaths="**/coverage.opencover.xml" \
           /d:sonar.cs.vstest.reportsPaths="**/*.trx"
       - name: Test
         run: |
           dotnet test specs/Qowaiv.Specs --configuration Release \
           --logger "trx;LogFileName=test-results.trx" \
-          /p:CollectCoverage=true \
-          /p:ThresholdType=branch \
-          /p:CoverletOutputFormat=opencover
+          --collect:"XPlat Code Coverage;Format=opencover"
       - name: Stop SonarCloud scanner
         if: success() || failure()
         env:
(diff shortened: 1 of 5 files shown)


Corniel:
Those multiple files, isn't that because these files are generated per test target?

.github/workflows/build-test.yaml
```diff
@@ -50,15 +50,13 @@ jobs:
...
+          /d:sonar.cs.opencover.reportsPaths="**/coverage.opencover.xml" \
           /d:sonar.cs.vstest.reportsPaths="**/*.trx"
       - name: Test
         run: |
           dotnet test specs/Qowaiv.Specs --configuration Release \
           --logger "trx;LogFileName=test-results.trx" \
-          /p:CollectCoverage=true \
-          /p:ThresholdType=branch \
-          /p:CoverletOutputFormat=opencover
+          --collect:"XPlat Code Coverage;Format=opencover"
```
Corniel, line 59: What is the meaning of this `XPlat Code Coverage`?

renzo-baasdam:
> Those multiple files, isn't that because these files are generated per test target?

No, there are actually 6 `coverage.opencover.xml` files compared to the expected 3 from test targets. This is something that has been notices before, but considered as [intended behavior](https://www.github.com/microsoft/vstest/issues/2334). Does not seem to cause problems for SonarCloud though.

image

.github/workflows/build-test.yaml
renzo-baasdam: Apparently it's the term to indicate the use `coverlet.collector`, see https://www.github.com/coverlet-coverage/coverlet/issues/1065.

Corniel (approved):
Lets gets this rolling.

PR #413, opened 2024-09-13
Fix code coverage pipeline

renzo-baasdam:
Wrongly assumed that `coverallsapp/github-action@master` supports `/**/` notation. Due to `coverlet.collector` saving `coverage.info` in a randomly generated Guid folder, it seems to be unavoidable to need to determine the path and save it for later, see https://www.github.com/coverallsapp/github-action/issues/36#issuecomment-741141560. Using a slightly different approach as `set-output` is [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -16,7 +16,12 @@ jobs:
           dotnet test ./specs/Qowaiv.Specs -f net8.0 \
           --configuration Release \
           --collect:"XPlat Code Coverage;Format=lcov"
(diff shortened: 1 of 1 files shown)


renzo-baasdam:
The pipeline run on `fix: lcov path usage` shows the success of the code coverage pipeline. Reverted to only run on `master` again in the commit after that.

Corniel (approved):
Let's try it out.

PR #417, opened 2024-09-28
Bug in wildcards with trailing chars at the end

Corniel:
While checking this code (for inspiration) I spotted this tiny bug.


specs/Qowaiv.Specs/Text/Wild_card_pattern_specs.cs
```diff
@@ -4,7 +4,97 @@ public class Is_match
...
+    [TestCase("ge?ks", "geeks")]
+    [TestCase("Qowaiv?", "Qowaiv")]
+    [TestCase("Qowaiv??", "Qowaiv")]
+    [TestCase("Qowaiv*", "Qowaiv")]
+    [TestCase("Qowaiv*?*", "Qowaiv")]
+    public void with_trailing(string pattern, string input) 
+        => WildcardPattern.IsMatch(pattern, input, WildcardPatternOptions.SingleOrTrailing, StringComparison.Ordinal)
+        .Should().BeTrue();
+
+    [TestCase("gee_ks", "geeks", true, WildcardPatternOptions.SingleOrTrailing | WildcardPatternOptions.SqlWildcards)]
```
renzo-baasdam, line 46: The `gee_ks` matches `geeks` here because of the ` WildcardPatternOptions.SingleOrTrailing`? Because I can't say that's what I would think of when seeing `Trailing`.

specs/Qowaiv.Specs/Text/Wild_card_pattern_specs.cs
```diff
@@ -4,7 +4,97 @@ public class Is_match
...
+                .Should().Be(isMatch);
+        }
+    }
+}
+
+public class Can_not_match
+{
+    [TestCase("gee?ks", "geeks", "No trailing")]
+    [TestCase("g*ks", "Geeks", "Capitalization")]
+    [TestCase("g*Ks", "geeks")]
```
renzo-baasdam, line 80: `because` in the second one is also "Capitalization"? (And also in the test below this one.)

renzo-baasdam (approved):

specs/Qowaiv.Specs/Text/Wild_card_pattern_specs.cs
Corniel: It felt double indeed.

Issue #418, opened 2024-10-09
Provide diagnostic contracts as code generator


Corniel:
To prevent runtime issues when applying reflection, it is better to provide `Qowaiv.Diganostics.Contracts` as a code generator package.

PR #419, opened 2024-10-09
Provide attributes as internals

Corniel:
While #418 has not been resolved, this reduces issues when using Qowaiv.

### DIFF
--- a/src/Qowaiv/Diagnostics/Contracts/CollectionMutationAttribute.cs
+++ b/src/Qowaiv/Diagnostics/Contracts/CollectionMutationAttribute.cs
@@ -1,0 +1,8 @@
+namespace Qowaiv.Diagnostics.Contracts;
+
+/// <summary>To mark a method explicitly as impure. Methods decorated with
+/// this attribute return info about (like, removal or addition was successful).
+/// </summary>
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+[Conditional("CONTRACTS_FULL")]
+internal sealed class CollectionMutationAttribute(string? justification = null) : ImpureAttribute(justification) { }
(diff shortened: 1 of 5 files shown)


CptWesley (approved):

src/Qowaiv/Diagnostics/Contracts/CollectionMutationAttribute.cs
```diff
@@ -0,0 +1,8 @@
+namespace Qowaiv.Diagnostics.Contracts;
+
+/// <summary>To mark a method explicitly as impure. Methods decorated with
```
Sjaaky, line 3: Comment is for impure instead of collectionmutation.

src/Qowaiv/Diagnostics/Contracts/FluentSyntaxAttribute.cs
```diff
@@ -0,0 +1,8 @@
+namespace Qowaiv.Diagnostics.Contracts;
+
+/// <summary>To mark a method explicitly as impure. Methods decorated with
```
Sjaaky, line 3: Comment is for impure instead of FluentSyntaxAttribute.

Sjaaky (approved):
I don't know if adding these attributes with the same class name and namespace in 2 different assemblies will give trouble in the future.

src/Qowaiv/Diagnostics/Contracts/CollectionMutationAttribute.cs
Corniel: That is not true. It states what to expect of the return value.

src/Qowaiv/Diagnostics/Contracts/FluentSyntaxAttribute.cs
Sjaaky: my bad. It is a special case of impure

Corniel: That is not true. It states that the return value is same instance as the input argument.

PR #422, opened 2024-10-10
Add examples to the intellisense of Percentage.Create(), .Percent() a…

Sjaaky:
…nd casts from and to numbers.

Add examples to commonly used method in order to avoid confusion.


Corniel (approved):
LGTM!

PR #426, opened 2024-10-24
Qowaiv.Diagnostics.Contracts Fix ambiguous references & support < C#10

### LINKED ISSUE #424: Qowaiv.Diagnostics.Contracts Ambiguous References
Whenever `Qowaiv.Diagnostics.Contracts` is used in a project that defines a type with a name that overlaps with any of the used types (e.g. `Conditional`) and this overlapping type is reachable by the name resolver (e.g. through `global using`), then the compiler is no longer able to determine which of the types to use. Therefore it would be better to use the full explicit names of the used types in the generated code.

### LINKED ISSUE #425: Qowaiv.Diagnostics.Contracts Not usable when using C# language 9 and lower
The generated code uses the "new" file-scoped namespace declarations introduced in C# 10. When a project is using a C# version that is lower than 10, this will cause compilation errors.

CptWesley:
Fixes #424 
Fixes #425

### DIFF
--- a/src/Qowaiv.Diagnostics.Contracts/.editorconfig
+++ b/src/Qowaiv.Diagnostics.Contracts/.editorconfig
@@ -1,0 +1,6 @@
+root = false
+
+# Static code analysis
+[*]
+dotnet_diagnostic.IDE0130.severity = none    # Folder structure does not match namespace. Justification: Files are only emited during code generation.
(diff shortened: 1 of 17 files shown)


Corniel (approved):
I once had the idea myself that this could be problematic for C# versions lacking file scoped namespace support. But than forgot about it. Happy you didn't! 

LGTM

PR #427, opened 2024-10-28
Non continous SVO's are preferably not serialized for their default values

Corniel:
I'm not certain if this `System.Text.Json` customization code should be in the Qowaiv core package in the first place, but besides, the use case is clear: For non-continuous SVO's (such as IBAN, emailadress, etc...) the JSON serialization value of the default/empty values is `null` (when `HasValue` is `false`).

Since .NET 8.0, it is possible to register modifiers to the `TypeInfoResolver`. One of them, is to assign `ShouldSerialize`. That option is used here. Using of this feature is straightforward:

``` C#
new JsonSerializerOptions()
{
    TypeInfoResolver = new DefaultJsonTypeInfoResolver()
    {
        Modifiers = { ModifyTypeInfo.IgnoreEmptySvos },
    },
};
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -703,7 +703,7 @@ based contract.
 * [Qowaiv.Json.Newtonsoft](https://www.nuget.org/packages/Qowaiv.Json.Newtonsoft/)  
 * [Qowaiv.Text.Json.Serialization](https://www.nuget.org/packages/Qowaiv.Text.Json.Serialization/)  
 
-For the .NET 6.0, and higher versions of the package, when using `System.Text.Json`,
+For .NET 6.0, and higher versions of the package, when using `System.Text.Json`,
 no custom serialization registration is required for Qowaiv SVO's: all have been
 decorated with the `[JsonConverter]` attribute.
 
--- a/README.md
+++ b/README.md
@@ -703,11 +703,28 @@ based contract.
 * [Qowaiv.Json.Newtonsoft](https://www.nuget.org/packages/Qowaiv.Json.Newtonsoft/)  
 * [Qowaiv.Text.Json.Serialization](https://www.nuget.org/packages/Qowaiv.Text.Json.Serialization/)  
 
-For the .NET 5.0, and higher versions of the package, when using `System.Text.Json`,
+For the .NET 6.0, and higher versions of the package, when using `System.Text.Json`,
 no custom serialization registration is required for Qowaiv SVO's: all have been
 decorated with the `[JsonConverter]` attribute.
 
-#### OpenAPI Specification
+#### Do not serialize empty SVO's
+Since .NET 8.0, it is possible to register modifiers for the `System.Text.Json.TypeInfoResolver`.
+This allows to change serialization behavior. On of the things to change is
+`ShouldSerialize(object model, object? prop)`. This can be useful if you do not
+want to serialize empty SVO's (such as `EmailAddress.Empty`). To get this
+(non-default) behavior you have to provide `JsonSerializerOptions` when serializing:
+
+``` C#
+var options = new JsonSerializerOptions()
+{
+    TypeInfoResolver = new DefaultJsonTypeInfoResolver()
+    {
+        Modifiers = { Qowaiv.Json.ModifyTypeInfo.IgnoreEmptySvos },
+    },
+};
+```
+
+### OpenAPI Specification
 The [OpenAPI Specification](https://swagger.io/docs/specification/about/)
 (formerly Swagger Specification) is an API description format for REST API's.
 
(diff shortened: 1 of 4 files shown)


Sjaaky:
looks good to me.

src/Qowaiv/Json/ModifyTypeInfo.cs
```diff
@@ -0,0 +1,38 @@
+#if NET8_0_OR_GREATER
```
CptWesley, line 1: `ShouldSerialize` is supported since .NET 6

Corniel: I could not find an entry point for .NET 6. Both `JsonTypeInfo` has no way to loop through properties (at least, I could not find it), and secondly, I could not find way to register `JsonTypeInfo` modifiers at the serializer options.

Given that .NET 6.0 will not longer be supported by Microsoft in 3 days time, I think it is safe to say I would not like to invest time i this, as it would require a separate implementation.

CptWesley: [https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.metadata.defaultjsontypeinforesolver.modifiers?view=net-6.0](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.metadata.defaultjsontypeinforesolver.modifiers?view=net-6.0) shows it being supported by `.NET 6`, but it mentions it is part of `System.Text.Json` `v8.0.5`, so perhaps it's only there if you target the `.NET 8` version of `System.Text.Json`.

CptWesley (approved):

README.md
```diff
@@ -703,11 +703,28 @@ based contract.
...
 no custom serialization registration is required for Qowaiv SVO's: all have been
 decorated with the `[JsonConverter]` attribute.
 
-#### OpenAPI Specification
+#### Do not serialize empty SVO's
+Since .NET 8.0, it is possible to register modifiers for the `System.Text.Json.TypeInfoResolver`.
+This allows to change serialization behavior. On of the things to change is
+`ShouldSerialize(object model, object? prop)`. This can be useful if you do not
+want to serialize empty SVO's (such as `EmailAddress.Empty`). To get this
+(non-default) behavior you have to provide `JsonSerializerOptions` when serializing:
```
LauraXiulan, line 715:
```suggestion
This allows you to change serialization behavior, for example by editing
`ShouldSerialize(object model, object? prop)`. This can be useful if you do not
want to serialize empty SVO's (such as `EmailAddress.Empty`). To get this
(non-default) behavior you have to provide `JsonSerializerOptions` when serializing:
```

specs/Qowaiv.Specs/Json/ModifyTypeInfo_specs.cs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    [Test]
+    public void serializes_non_empty_svos()
+    {
+        var json = JsonSerializer.Serialize(new WithNonContinuousSvo { Iban = Svo.Iban }, Options);
+        json.Should().Be(@"{""Iban"":""NL20INGB0001234567""}");
+    }
+
+    private sealed class WithNonContinuousSvo { public InternationalBankAccountNumber Iban { get; init; } }
```
LauraXiulan, line 33: What do you mean with NonContinuous?

README.md
```diff
@@ -703,11 +703,28 @@ based contract.
 * [Qowaiv.Json.Newtonsoft](https://www.nuget.org/packages/Qowaiv.Json.Newtonsoft/)  
 * [Qowaiv.Text.Json.Serialization](https://www.nuget.org/packages/Qowaiv.Text.Json.Serialization/)  
 
-For the .NET 5.0, and higher versions of the package, when using `System.Text.Json`,
+For the .NET 6.0, and higher versions of the package, when using `System.Text.Json`,
```
LauraXiulan, line 706:
```suggestion
For .NET 6.0, and higher versions of the package, when using `System.Text.Json`,
```

src/Qowaiv/Json/ModifyTypeInfo.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        Guard.NotNull(info);
+
+        foreach (var prop in info.Properties.Where(IsApplicable))
+        {
+            prop.ShouldSerialize = ShouldSerialize(Activator.CreateInstance(prop.PropertyType)!);
+        }
+    }
+
+    /// <summary>True when the propValue is different to the empty/default value.</summary>
```
LauraXiulan, line 24:
```suggestion
    /// <summary>True when the propValue is not equal to the empty/default value.</summary>
```

specs/Qowaiv.Specs/Json/ModifyTypeInfo_specs.cs
Corniel: The mathematical definition: [](https://en.wikipedia.org/wiki/Continuous_function).

SO types like decimals, percentages, amounts, integers are continuous, but IBAN's, Email Addresses e.o. are not: they have a well defined empty state.

LauraXiulan (approved):

PR #431, opened 2024-11-12
Use new .NET 9 Lock type

CptWesley:
See: https://thetechplatform.medium.com/system-threading-lock-a-new-thread-synchronization-in-net-9-cf850339b2dc

### DIFF
--- a/src/Qowaiv/Extensions/System.Type.cs
+++ b/src/Qowaiv/Extensions/System.Type.cs
@@ -1,11 +1,20 @@
 using Qowaiv;
 using System.Runtime.CompilerServices;
+using System.Threading;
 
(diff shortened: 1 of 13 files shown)


src/Qowaiv/Extensions/System.Type.cs
```diff
@@ -1,11 +1,20 @@
 using Qowaiv;
 using System.Runtime.CompilerServices;
+using System.Threading;
```
Corniel, line 3: Move this one to the global usings

Corniel (approved):
Please update the release notes (ToBeReleased) for this. For the rest: LGTM!

PR #436, opened 2024-11-15
Introduction FileInfo.Lock() to lock files during a test

Corniel:
To test IO related unhappy flows, it can be useful to create a temporary lock on a file. This can be done as follows:

``` C#

var file = new FileInfo("somefile.txt");

using (var @lock = file.Lock())
{
     file.OpenRead(); // Throws IOException.
}
```

### DIFF
--- a/specs/Qowaiv.Specs/Clock_specs.cs
+++ b/specs/Qowaiv.Specs/Clock_specs.cs
@@ -103,24 +103,6 @@ public class For_current_execution_context_and_scope
         Clock.UtcNow().Should().NotBe(Svo.DateTime);
         Clock.TimeZone.Should().NotBe(Svo.TimeZone);
     }
-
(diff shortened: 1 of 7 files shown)


specs/Qowaiv.Specs/Clock_specs.cs
```diff
@@ -103,24 +103,6 @@ public void UtcNow_and_TimeZone_can_be_set_together()
         Clock.UtcNow().Should().NotBe(Svo.DateTime);
         Clock.TimeZone.Should().NotBe(Svo.TimeZone);
     }
-
```
Sjaaky, line 106: Seems unrelated

Corniel: Was a flacky test. But, yes unrelated.

PR #441, opened 2024-12-17
Add deconstructors for relevent types

Corniel:
The option to [deconstruct types](https://learn.microsoft.com/dotnet/csharp/fundamentals/functional/deconstruct) exist since C# 7, so it was about time to add the support to SVO's that benefit from it:
- [x] Date
- [x] DateSpan
- [x] Fraction
- [x] Money
- [x] MonthSpan
- [x] WeekDate
- [x] YearMonth

It also has been added to `FormattingAgurments`, an helper to combine the arguments provided to apply `.ToString(string format, IFormatProvider formatProvider)`, defined by `IFormattable`.

### DIFF
--- a/specs/Qowaiv.Specs/Date_span_specs.cs
+++ b/specs/Qowaiv.Specs/Date_span_specs.cs
@@ -292,3 +292,15 @@ public class Throws_when
             .Should().Throw<ArgumentOutOfRangeException>()
             .WithMessage("The specified years, months and days results in an un-representable DateSpan.");
 }
+
+public class Can_be_deconstructed
+{
+    [Test]
+    public void in_years_months_and_days_part()
+    {
+        var (years, months, days) = Svo.DateSpan;
+        years.Should().Be(10);
+        months.Should().Be(3);
+        days.Should().Be(-5);
+    }
+}
--- a/specs/Qowaiv.Specs/Week_date_specs.cs
+++ b/specs/Qowaiv.Specs/Week_date_specs.cs
@@ -212,3 +212,15 @@ public class Casts
     }
 }
 #endif
+
+public class Can_be_deconstructed
+{
+    [Test]
+    public void int_year_week_and_day_part()
+    {
+        var (year, week, day) = Svo.WeekDate;
+        year.Should().Be(2017);
+        week.Should().Be(23);
+        day.Should().Be(7);
+    }
+}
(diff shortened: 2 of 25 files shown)


CptWesley:
Not sure if useful information for you right now, but it is also possible to have multiple deconstructors for the same type, in order to allow for deconstructing in pattern matching. e.g.:

```csharp
var x = ym switch
{
    YearMonth("2024-01") => "foo",
    YearMonth(string str) => str,
    YearMonth(2024, 1) => "bar",
    YearMonth(int y, int m) => "baz",
    _ => string.Empty,
};
```

But this might cause ambiguity conflicts when deconstructing in the way you are currently doing in the tests when the number of arguments is the same.

CptWesley (approved):

Corniel:
> Not sure if useful information for you right now, but it is also possible to have multiple deconstructors for the same type, in order to allow for deconstructing in pattern matching. e.g. (..)

I'd tried to stick with the same guidelines as MS did in .NET. Do you have a concrete suggestion of a deconstruct you want add?

CptWesley:
I was initially thinking to be able to match on the string version, but it gets finicky with different cultures, so perhaps it's for the better to exclude it

specs/Qowaiv.Specs/Date_span_specs.cs
```diff
@@ -292,3 +292,15 @@ public void Ctor_arguments_are_out_of_range()
...
+
+public class Can_be_deconstructed
+{
+    [Test]
+    public void in_years_months_and_days_part()
+    {
+        var (years, months, days) = Svo.DateSpan;
+        years.Should().Be(10);
+        months.Should().Be(3);
+        days.Should().Be(-5);
```
LauraXiulan, line 304: -5?

specs/Qowaiv.Specs/Week_date_specs.cs
```diff
@@ -212,3 +212,15 @@ public void implicitly_from_DateOnly()
     }
 }
 #endif
+
+public class Can_be_deconstructed
+{
+    [Test]
+    public void int_year_week_and_day_part()
```
LauraXiulan, line 219:
```suggestion
    public void in_year_week_and_day_part()
```

LauraXiulan (approved):

PR #442, opened 2024-12-17
Tweak Year-month Open API definition pattern

Corniel:
That pattern for Year-month as specified by the Open API attribute should not only allow `2017-06`, but also `2017-6`, even though Qowaiv will seralize with the leading zero.

### DIFF
--- a/README.md
+++ b/README.md
@@ -836,7 +836,7 @@ and if the data type is nullable, all when applicable.
     "example": "2017-06",
     "type": "string",
     "format": "year-month",
-    "pattern": "[0-9]{4}-[01][0-9]",
+    "pattern": "[0-9]{4}-(0?[1-9]|1[0-2])",
     "nullable": false
   },
   "YesNo": {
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -836,7 +836,7 @@ and if the data type is nullable, all when applicable.
     "example": "2017-06",
     "type": "string",
     "format": "year-month",
-    "pattern": "[0-9]{4}-[01][0-9]",
+    "pattern": "[0-9]{4}-(0?[1-9]|1[0-2])",
```
CptWesley, line 839: This does not match `10` correctly.

CptWesley: Actually, it will work fine if the entire input has to be consumed by the match

CptWesley (approved):

README.md
Corniel: But [0-9]{4}-(0?[1-9]|**1[0-2]**) does, right?

CptWesley: A plain regex matches left-to-right eagerly, so `2024-10` would match `2024-1`, because `1` matches `0?[1-9]`. But if the regex requires the full input to be consumed. If the `2024-1` match was rejected because it does not include the `0`, then it would still end up matching `2024-10` with some extra backtracking in the end. Swapping the order to `(1[0-2]|0?[1-9])` would make it work in all cases, but it's not that relevant

PR #444, opened 2025-01-03
PostalCode.Empty is valid for countries without a postal code system

### LINKED ISSUE #443: Postalcode properly use nullable values
The postalcode code seems to be a mix of nullable and non-nullable conventions. Since the existing code shows m_Value can be null. It looks better to me if we remove the unknown postal code and check for Mvalue is null instead. another option is to do the is null check in Unknown(). This will be a breaking change because there might be databases storing the "ZZZZZZZZZ" unknown value.

Continuing the use of nullable values. Isvalid(Country) should be true if the postalcode is null and the country code doesn't have a postal code system.

Corniel:
For countries without a postal code system, `PostalCode.Empty` should be valid (and all other values should not be valid, what was already the case).

Closes #443

### DIFF
--- a/src/Qowaiv/Globalization/PostalCodeCountryInfo.cs
+++ b/src/Qowaiv/Globalization/PostalCodeCountryInfo.cs
@@ -49,9 +49,10 @@ public sealed partial class PostalCodeCountryInfo
     /// </remarks>
     [Pure]
     public bool IsValid(string? postalcode)
-        => !string.IsNullOrEmpty(postalcode)
-        && ValidationPattern is { }
-        && postalcode.Unify().Matches(ValidationPattern);
+        => HasPostalCode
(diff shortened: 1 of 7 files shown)


src/Qowaiv/Globalization/PostalCodeCountryInfo.cs
```diff
@@ -49,9 +49,10 @@ private PostalCodeCountryInfo(
     /// </remarks>
```
Ard2025, line 49: The documentation needs an update for the changed functionality, what do you think of the following?

``` 
    /// <remarks>
    /// Returns false if the postal code is not empty and the country does not have postal codes.
    /// </remarks>```

Corniel: Well spotted. I've updated the `<remarks>`.

PR #446, opened 2025-01-03
Trying to fix build issue with an icu version override

Gemberkoekje:


.github/workflows/build-test.yaml
```diff
@@ -1,10 +1,14 @@
 name: Build
 on: [push]
+env:
+  CLR_ICU_VERSION_OVERRIDE: 74
 jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
     steps:
+      - name: Echo versions
```
Corniel: Is this required, or just for debugging purposes?

.github/workflows/build-test.yaml
```diff
@@ -1,10 +1,14 @@
 name: Build
 on: [push]
+env:
```
Corniel, line 3: Might be worth to document why this is needed?!

Corniel:
Note, when you rebase/FI, please enable the netcore3.1 target again for the unit tests.

Corniel:
Solved With #452

PR #447, opened 2025-01-03
Update documentation of IUnknown interface

Corniel:
It turned out that the documentation could be better. (See #444)

### DIFF
--- a/src/Qowaiv/Contracts/IUnknown.cs
+++ b/src/Qowaiv/Contracts/IUnknown.cs
@@ -1,8 +1,12 @@
 namespace Qowaiv;
 
 /// <summary>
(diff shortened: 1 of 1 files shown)


src/Qowaiv/Contracts/IUnknown.cs
```diff
@@ -1,8 +1,12 @@
 namespace Qowaiv;
 
 /// <summary>
-/// Defines that the default of an SVO represents the empty/not set state.
+/// Defines that an SVO has both an empty, a known, and an unknown state.
```
LauraXiulan, line 4:
```suggestion
/// Defines that a SVO has both an empty, a known, and an unknown state.
```

PR #448, opened 2025-01-05
Use Random.Shared in DecimalMath

Corniel:
It is preferred to use built-in functionality of .NET.

### DIFF
--- a/src/Qowaiv/Mathematics/DecimalMath.cs
+++ b/src/Qowaiv/Mathematics/DecimalMath.cs
@@ -151,6 +151,7 @@ internal static class DecimalMath
     /// creates a new instance if required.
     /// </remarks>
     [Pure]
+#if NETSTANDARD2_0
     private static Random Random()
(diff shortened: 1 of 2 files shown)


src/Qowaiv/Mathematics/DecimalMath.cs
```diff
@@ -159,4 +160,7 @@ private static Random Random()
 
     [ThreadStatic]
     private static Random? _rnd;
+#else
+    private static Random Random() => System.Random.Shared;
```
CptWesley, line 164: Perhaps a good idea to add a comment here noting that `System.Random.Shared` is thread-safe

CptWesley (approved):
LGTM

src/Qowaiv/Mathematics/DecimalMath.cs
Corniel: Added.

PR #452, opened 2025-01-24
Test .NET standard 2.0 via .NET 5.0

Corniel:
We found the hard way that we not longer could run the builds on .NET Core 3.1 (see #446 ). By using .NET 5.0 instead of trying to do really complex stuff, we now (since the .NET 5.0 targets have been dropped #359) can use it to test the .NET standard 2.0 target.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -17,10 +17,10 @@ jobs:
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: |
-            3.1.x
-            6.0.x
-            8.0.x
-            9.0.x
+            5.x
+            6.x
+            8.x
+            9.x
       - name: Cache SonarCloud packages
         uses: actions/cache@v4
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test.yaml
```diff
@@ -17,10 +17,14 @@ jobs:
...
-            8.0.x
-            9.0.x
+            5.x
+            6.x
+            8.x
+            9.x
+      - name: Install OpenSSL 1.1 # Needed for dotnet 5.x
+        run: |
+          wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2.23_amd64.deb
+          sudo dpkg -i libssl1.1_1.1.1f-1ubuntu2.23_amd64.deb
```
renzo-baasdam, line 27: Looks like .NET 5.0 requires libssl 1.x, while the build-agent `ubuntu-latest` has 3.0. So manually installing libssl 1.1 seems to work. (Also see [this](https://www.github.com/dotnet/core/issues/4749))

renzo-baasdam (approved):

PR #457, opened 2025-02-08
Improve generated attributes

CptWesley:
- Added `// <auto-generated/>` because pragma disable somehow doesn't work for all analyzers (looking at you StyleCop).
- Added `#nullable enable` because nullability analysis is disabled by default in generated code (since it might not match the project settings that it is generated into). This should make the nullability analysis work properly again for consuming apps.
- Prefixed all type named with `global::` to ensure there can't be any ambiguity when someone for some reason has decided to create a conflicting namespace with `Qowaiv` or `System`.


Corniel (approved):
LGTM!

PR #458, opened 2025-02-08
Use [SkipLocalsInit]

Corniel:
More info
* [learn.microsoft.com](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.skiplocalsinitattribute)
* [meziantou.net](https://www.meziantou.net/csharp-9-improve-performance-using-skiplocalsinit.htm)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <IsPublishable>false</IsPublishable>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <OutputType>library</OutputType>
   </PropertyGroup>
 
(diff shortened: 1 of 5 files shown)


CptWesley:
Did you verify that this doesn't have any unforeseen consequences?

Corniel:
> Did you verify that this doesn't have any unforeseen consequences?

Not yet. I'm also not sure I'm in favor of the change yet. The gain is tiny, and the risks might be tiny too, but I have no understanding yet which they might be in our case. Until now, we never needed the `AllowUnsafeBlocks` flag (we do depend on unsafe code blocks, for XML deseralization, but that is external code from MS).

Corniel:
I found this interesting discussion: https://github.com/dotnet/aspnetcore/issues/26586

Not sure about the implications yet dough.

Corniel:
The one place that uses `stackalloc` seems to improve slightly indeed:

| Uuid.ToString()      |   New |   Old |
|--------------------- |------:|------:|
| GUID (Guid.ToString) |  1.00 |  1.00 |
| Base64               |  1.45 |  1.57 |
| GUID                 |  4.04 |  3.85 |
| Convert.ToBase64     |  4.75 |  4.57 |

CptWesley (approved):
LGTM

PR #460, opened 2025-02-14
Email parser reimplemented

Corniel:
After implementing email parsing in [TypeScript](https://github.com/Qowaiv/qowaiv-js/pull/22) I felt inspired to improve om the current implementation. By using `ReadOnlySpan<char>` and a `Span<char>` with a `stackloc`, and simplifying on the structure on the parser (mostly by being a bit slower an a Display name within quotes in front of an email address with a quoted local part) I was able to reduce the amount of lines needed, and made  2.37 times faster too.

| Parse | Mean      | Ratio |
|---------- |----------:|------:|
| CharBuffer | 176.39 us |  2.37 |
| Span |  74.35 us |  1.00 |

I'm happy with the results.

Note that a `ref` struct is use to keep the spans. As a result nullables can not be used, what would have made some code easier to grasp:

```` C#
=> Mailto()?.Local()?.Domain();
````

end 

``` C#
=> DomainRegular() ?? DomainIp():
```

But the use of spans is more important ( I guess).


src/Qowaiv/Email.cs
```diff
@@ -0,0 +1,301 @@
...
+    private const int TotalLength = 254;
+    private const char At = '@';
+    private const char Dot = '.';
+    private const char Dash = '-';
+    private const char Quote = '"';
+
+    [Pure]
+    public static string? Parse(ReadOnlySpan<char> str)
+    {
+        Span<char> buffer = stackalloc char[TotalLength + 1];
```
CptWesley, line 19: Why the +1?

Corniel: The `Next()` method returns the next `char`, `index`, and `length`, after writing the `char` in the `Buffer`. Only afterwards it determined if the length has been exceeded. So without the `+ 1` an `IndexOutOfRange` exception would occur. Do you think I should add a comment?

CptWesley: Perhaps. I was trying to make sure nothing breaks due to the recently added `SkipLocalsInit`

Corniel: A part of the speed improvements are due to the `[SkipLocalsInit]`. The code always keeps track on what has been written (you can see the `[SkipLocalsInit]` in action while debugging). The previous implementation used a fully allocated char array.

CptWesley (approved):

PR #464, opened 2025-03-06
Drop CharBuffer

Corniel:
Once, I introduced a `CharBuffer` to have a `StringBuilder` like unit to tweaks string input. With the adoption of `Span<char>` this code has become obsolete. With this PR, I removed the latest code that used it (`ToNonDiacrtic()` and `Unify()`).


src/Qowaiv/Qowaiv.csproj
```diff
@@ -46,6 +46,7 @@ ToBeReleased:
 - Use [SkipLocalsInit] (speed improvement). #458
 - Email address topdomain can not be longer than 63. (fix) #459
 - Rewritten email parser 2.4 times faster. #460
+- Drop CharBuffer in favor of Span<char> based implementations.
```
renzo-baasdam, line 49:
```suggestion
- Drop CharBuffer in favor of Span<char> based implementations. #464 
```

src/Qowaiv/Extensions/System.String.cs
```diff
@@ -70,7 +70,17 @@ internal static string WithDefault(this string? str, string @default = "")
         => str is { Length: > 0 } ? str : @default;
 
     [Pure]
-    internal static string Unify(this string? str) => str.Buffer().Unify();
+    internal static string Unify(this string? str)
+    {
+        if (str is { Length: > 0 })
+        {
+            ReadOnlySpan<char> span = str;
+            Span<char> buffer = stackalloc char[span.BufferSize()];
```
renzo-baasdam, line 78: `stackalloc` for performance reasons?

renzo-baasdam (approved):

src/Qowaiv/Text/CharBuffer.cs
```diff
@@ -1,57 +1,112 @@
...
-    [Pure]
-    [ExcludeFromCodeCoverage]
-    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+    private static readonly Dictionary<char, string> DiacriticLookup = new()
+    {
+        { 'Æ', "AE" },
+        { 'Ǽ', "AE" },
+        { 'æ', "ae" },
+        { 'ǽ', "ae" },
+        { 'ß', "sz" },
```
CptWesley, line 91: Shouldn't this be `ss` rather than `sz`?

src/Qowaiv/Text/CharBuffer.cs
```diff
@@ -1,57 +1,112 @@
...
-    public char First() => buffer[start];
+        foreach (var ch in span)
+        {
+            if (ignore.Contains(ch))
+            {
+                buffer[length++] = ch;
+            }
+            else
+            {
+                var relace = DiacriticSearch.IndexOf(ch);
```
CptWesley, line 62: Performance-wise it would probably be better to use a `FrozenSet` (or for targets that don't support frozen collections, use a regular old `HashSet`). I understand it's easier to just define this set as a `string`, but since a `string` is an `IEnumerable<char>` anyway, you could just do `.ToFrozenSet()` or `.ToHashSet()` on the string.

CptWesley: Edit: or use a `FrozenDictionary` instead to find the replacement.

src/Qowaiv/Text/CharBuffer.cs
```diff
@@ -1,57 +1,112 @@
...
-    [Pure]
-    public IEnumerator<char> GetEnumerator() => buffer.Skip(start).Take(Length).GetEnumerator();
+    private const string DiacriticSearch = /*..*/ "ÀÁÂÃÄÅĀĂĄǍǺàáâãäåāăąǎǻÇĆĈĊČƠçćĉċčơÐĎďđÈÉÊËĒĔĖĘĚèéêëēĕėęěÌÍÎÏĨĪĬĮİǏìíîïıĩīĭįǐĴĵĜĞĠĢĝğġģĤĦĥħĶķĸĹĻĽĿŁĺļľŀłÑŃŅŇŊñńņňŉŋÒÓÔÕÖØŌŎŐǑǾðòóôõöøōŏőǒǿŔŖŘŕŗřŚŜŞŠśŝşšŢŤŦţťŧÙÚÛÜŨŪŬŮŰŲƯǓǕǗǙǛùúûüũūŭůűųưǔǖǘǚǜŴŵÝŶŸýÿŷŹŻŽźżž";
+    private const string DiacriticReplace = /*.*/ "AAAAAAAAAAAaaaaaaaaaaaCCCCCCccccccDDddEEEEEEEEEeeeeeeeeeIIIIIIIIIIiiiiiiiiiiJjGGGGggggHHhhKkkLLLLLlllllNNNNNnnnnnnOOOOOOOOOOOooooooooooooRRRrrrSSSSssssTTTtttUUUUUUUUUUUUUUUUuuuuuuuuuuuuuuuuWwYYYyyyZZZzzz";
 
-    /// <inheritdoc />
-    [Pure]
-    [ExcludeFromCodeCoverage]
-    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+    private static readonly Dictionary<char, string> DiacriticLookup = new()
```
CptWesley, line 85: Perhaps conditionally use a `FrozenDictionary` for better performance if the target allows.

src/Qowaiv/Text/CharBuffer.cs
Corniel: Not according to [Wikipedia](https://en.wikipedia.org/wiki/%C3%9F)

src/Qowaiv/Text/CharBuffer.cs
Corniel: But based on the index we get the replacement. And a set will not tell you its order?!

src/Qowaiv/Text/CharBuffer.cs
Corniel: Yep, I already thought of this. There are more places where that could (slightly) improve the performance. I created a separate task for it: #465

As this is about getting rid of the old CharBuffer, I'll not touch this yet.

src/Qowaiv/Text/CharBuffer.cs
CptWesley: Hence the suggestion to use a dictionary

Corniel: @CptWesley  I've applied the suggestion of using a dictionary. Converting that one to a frozen dictionary will be handled when #465 is solved.

PR #471, opened 2025-03-11
Collection initalization simplifications

Corniel:


Corniel:
This is not good: 
> Expected EmailAddress.Parse(@"mailto:""quoted""@qowaiv.org") to be equal to "quoted"ஸqowaiv.org, but found "quoted"师qowaiv.org.

PR #473, opened 2025-03-18
Hide Guard methods from stack trace

Corniel:
I stumbled on the `[StackTraceHidden]` attribute (by accident) I thought, that would be nice to remove the `Guard` methods from the stack trace. It turns out, I'm not the only one who thinks so, as I found this blog: [Hide a method from the stack trace](https://makolyte.com/csharp-exclude-exception-throw-helper-methods-from-the-stack-trace)


Sjaaky (approved):
Lgtm

PR #475, opened 2025-03-21
Decorate all code with [ExcludeFromCodeCoverage]

Corniel:
Although all code is generated with `// <auto-generated/>`  It turned out when I run a unit project using these attributes, it affected the code coverage nevertheless. Hence this little change.


CptWesley (approved):
LGTM

PR #480, opened 2025-04-25
Reduce required C# language version from 12 to 10

Corniel:
Primary constructors are only available since C# 12. By not using them, the language requirements is loosen.


Sjaaky:
Lgtm

src/Qowaiv.Diagnostics.Contracts/Qowaiv.Diagnostics.Contracts.csproj
```diff
@@ -12,8 +12,11 @@
     <Version>2.0.5</Version>
```
Gemberkoekje, line 12: Does version need to be updated as well?

Corniel:
@CptWesley That has already been done, hence the reason I found the issue in the first place.

src/Qowaiv.Diagnostics.Contracts/Qowaiv.Diagnostics.Contracts.csproj
Corniel: Oops

PR #483, opened 2025-05-06
Generate ID's

Corniel:
Comparable for #478 , but not for generic ID's, as a replacement for `Id<TBehavior>`.


src/Qowaiv/Customization/CustomBehavior.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// <inheritdoc />
+    [Pure]
+    public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value)
+        => TryTransform(value?.ToString(), culture, out var transformed)
+        ? transformed
+        : throw InvalidFormat(value?.ToString(), culture);
+
+    /// <summary>Updates <see cref="SvoType"/>.</summary>
+    [FluentSyntax]
+    public TypeConverter WithSvoType(Type type)
```
JurJean, line 36: Perhaps make an AdHoc behavior? Because otherwise this method exists on non-adhoc behaviors, right?

Corniel: What do you mean? This one is needed to create a working instance of the actual `TypeConverer` used by the ID.

PR #493, opened 2025-05-22
Use pattern matching instead of if else if constructs

Corniel:
There are quite some `if else if else` blocks in this codebase. They have been written before the modern pattern matching switch statements exists. Now it does, writing it in such matter further improves the readabillity.

### DIFF
--- a/src/Qowaiv/Financial/BusinessIdentifierCode.cs
+++ b/src/Qowaiv/Financial/BusinessIdentifierCode.cs
@@ -91,15 +91,13 @@ public readonly partial struct BusinessIdentifierCode : IXmlSerializable, IForma
     /// The format provider.
     /// </param>
     [Pure]
-    public string ToString(string? format, IFormatProvider? formatProvider)
+    public string ToString(string? format, IFormatProvider? formatProvider) => m_Value switch
     {
-        if (StringFormatter.TryApplyCustomFormatter(format, this, formatProvider, out string formatted))
-        {
-            return formatted;
-        }
-        else if (IsUnknown()) return "?";
-        else return m_Value ?? string.Empty;
-    }
+        _ when StringFormatter.TryApplyCustomFormatter(format, this, formatProvider, out string formatted) => formatted,
+        _ when !HasValue => string.Empty,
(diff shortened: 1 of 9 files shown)


src/Qowaiv/Financial/BusinessIdentifierCode.cs
```diff
@@ -91,15 +91,13 @@ public string Location
...
     {
-        if (StringFormatter.TryApplyCustomFormatter(format, this, formatProvider, out string formatted))
-        {
-            return formatted;
-        }
-        else if (IsUnknown()) return "?";
-        else return m_Value ?? string.Empty;
-    }
+        _ when StringFormatter.TryApplyCustomFormatter(format, this, formatProvider, out string formatted) => formatted,
+        _ when !HasValue => string.Empty,
```
Silvex78, line 97: HasValue and IsKnown are switched from the original?

src/Qowaiv/HouseNumber.cs
```diff
@@ -94,16 +94,13 @@ public int Length
...
-        if (StringFormatter.TryApplyCustomFormatter(format, this, formatProvider, out string formatted))
-        {
-            return formatted;
-        }
-        else if (IsUnknown()) return "?";
-        else if (IsEmpty()) return string.Empty;
-        else return m_Value.ToString(format, formatProvider);
-    }
+        _ when StringFormatter.TryApplyCustomFormatter(format, this, formatProvider, out string formatted)=> formatted,
+        _ when !HasValue => string.Empty,
```
Silvex78, line 100: Unknown and Empty switched here as well?

src/Qowaiv/Financial/BusinessIdentifierCode.cs
Corniel: In the order of processing yes, but the outcome is the same (the tests proof that).

src/Qowaiv/HouseNumber.cs
Corniel: Same answer,

Silvex78 (approved):

PR #526, opened 2025-11-01
Increase performance of svo/id generators

CptWesley:
There are two issues in current implementation:
1. ImmutableArray doesn't have "deep equality", so each time the generator runs (every keystroke) causes the generator to think the input changed. (see [here](https://andrewlock.net/creating-a-source-generator-part-9-avoiding-performance-pitfalls-in-incremental-generators/#4-watch-out-for-collection-types)).
2. **IF** ImmutableArray would be replaced with something that is actually equatable for caching purposes, the code for all SVO/IDs would be regenerated whenever any SVO/ID definition would change or whenever one would be added/removed

This change causes the caching to work properly and to be done for each individual SVO/ID, rather than all of them at once.

For my personal source generators, I usually at tests that:
1. Validate a snapshot of the output.
2. Check that for 2 consecutive runs on the same input code, the cached results were used, rather than regenerated.

If you want, I can add those kind of tests to the Qowaiv source generators as well (in the upcoming days).

### DIFF
--- a/src/Qowaiv.CodeGeneration.SingleValueObjects/BaseGenerator.cs
+++ b/src/Qowaiv.CodeGeneration.SingleValueObjects/BaseGenerator.cs
@@ -1,5 +1,4 @@
 using Microsoft.CodeAnalysis;
-using System.Collections.Immutable;
 using System.Threading;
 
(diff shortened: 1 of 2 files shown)


Corniel (approved):
LGTM!

PR #573, opened 2026-06-20
Add obfuscation to IBAN

Corniel:
There are contexts where it is better to not show the full IBAN, to prevent exposure of sensitive data. With the format `"O"` (or `"o"`), an IBAN is obfuscated: `NL20XXXXXXXXXX4567`.

### DIFF
--- a/src/Qowaiv/Financial/InternationalBankAccountNumber.cs
+++ b/src/Qowaiv/Financial/InternationalBankAccountNumber.cs
@@ -111,7 +111,7 @@ public readonly partial struct InternationalBankAccountNumber : IXmlSerializable
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


src/Qowaiv/Financial/InternationalBankAccountNumber.cs
```diff
@@ -111,6 +110,16 @@ public string HumanReadable(char space)
         }
     }
 
+    /// <summary>
+    /// Masks sensitive parts of an IBANto protect user privacy. (eg: NL20XXXXXXXXXX4567).</summary>
```
LauraXiulan:
```suggestion
    /// Masks sensitive parts of an IBAN to protect user privacy (eg: NL20XXXXXXXXXX4567).</summary>
```

LauraXiulan (approved):
One comment

PR #574, opened 2026-06-22
Obfuscate all but last 3 character

Corniel:
It might be even better to obfuscate an extra character.


Sjaaky (approved):
lgtm

ArnoudTjip:
Obfuscation of a Dutch IBAN is slightly safer when it only displays the last 3 digits. Due to the '11-proef' requirements, with 4 digits the range of possible correct IBAN accounts is often below 50, and sometimes below 20. As obfuscating the data gives users a sense of security (for example: they are more willing to share screenshots), moving to 3 digits is safer; and still serves the purpose of allowing users to recognize their own IBAN.

ArnoudTjip (approved):

PR #579, opened 2026-07-03
Simplify IBAN Parsing

Corniel:
Inspired by other work I was doing, I decided to try if I could squeeze some nano-seconds out using `Span<char>` and `ReadOnlySpan<char>`. I also decided to simplify the way of choosing BBAN info.

Originally, I considered implementing extra BBAN specific logic (like the Dutch 11-proof), but as that could lead to false positives when those extra constraints are dropped, a mechanism with different BBAN parsers, being chosen based on the country code became clumsy.

Furthermore, I decided to first try to parse the IBAN assuming it would be unformulated, as that is the most common format to deal with. It comes with a small penalty when parsing formatted IBAN's, and some code duplication, but a significant speed improvement for parsing the default ones, which made it worth it.

| Method   | Categories  | Mean      | Ratio | Gen0    | Gen1   | Allocated | Alloc Ratio |
|--------- |------------ |----------:|------:|--------:|-------:|----------:|------------:|
| Qowaiv   | Unformatted |  92.95 ns |  1.00 |  5.7373 | 1.3428 |   70.9 KB |        1.00 |
| Iban.NET | Unformatted | 271.90 ns |  2.93 | 12.2070 | 3.9063 |  155.3 KB |        2.19 |
|          |             |           |       |         |        |           |             |
| Qowaiv   | Formatted   | 130.85 ns |  1.00 |  5.6152 | 1.2207 |   70.9 KB |        1.00 |
| Iban.NET | Formatted   | 280.17 ns |  2.14 | 12.2070 | 3.9063 |  155.3 KB |        2.19 |


src/Qowaiv/Financial/IbanParser.cs
```diff
@@ -1,13 +1,9 @@
-using System.Runtime.CompilerServices;
-
 namespace Qowaiv.Financial;
```
Corniel, line 1:
```suggestion
// <auto-generated/>
namespace Qowaiv.Financial;
```

Gemberkoekje (approved):

Sjaaky (approved):
lgtm

src/Qowaiv/Financial/IbanParser.cs
```diff
@@ -16,63 +12,194 @@ internal static partial class IbanParser
...
+            ? iban
+            : null;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        static bool IsMarkup(char ch)
+            => ASCII.IsAscii(ch)
+            ? ASCII.IsMarkup(ch)
+            : char.IsWhiteSpace(ch);
+
+        /// <summary>Starts with "(IBAN)".</summary>
```
Corniel, line 122:
```suggestion
        // Starts with "(IBAN)".
```

src/Qowaiv/Financial/IbanParser.cs
```diff
@@ -16,63 +12,194 @@ internal static partial class IbanParser
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        static bool Has_IBAN_Prefix(ReadOnlySpan<char> reader)
+            => reader.StartsWith("(IBAN)", StringComparison.OrdinalIgnoreCase);
+
+        /// <summary>Starts with "IBAN " or "IBAN:".</summary>
```
Corniel, line 127:
```suggestion
        // Starts with "IBAN " or "IBAN:".
```

PR #582, opened 2026-07-09
Improve debugger display

Corniel:
Set all `[DebuggerDisplay("{DebuggerDisplay,nq}")]` to drop quotes from debugger display.

This changes the behavior of the IDE:

Here the  SVO value is now displayed as  `AEGONL2UXXX` , instead of  `"AEGONL2UXXX"`.


src/Qowaiv/Financial/BusinessIdentifierCode.cs
```diff
@@ -19,7 +19,7 @@ namespace Qowaiv.Financial;
 /// messages between banks. The codes can sometimes be found on account
 /// statements.
 /// </remarks>
-[DebuggerDisplay("{DebuggerDisplay}")]
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
```
Sjaaky, line 22: The BIC is based on a string. With nq it will be displayed without quotes, is that better or just confusing?
Same question for other string based svo's.

Sjaaky (approved):
lgtm. Small question.

src/Qowaiv/Financial/BusinessIdentifierCode.cs
Corniel: What would quotes add to the `AEGONL2UXXX` value here according to you?


I would argue that it now better hints that the `svo` is not a `string`, but something else, which it is. It is an identifier that can be represented by a set of charterers, constrainted by its domain logic.

PR #583, opened 2026-07-09
Percentage JSON serialization improvements

Corniel:
.NET support writing and reading from UTF8 streams for some time, but none of the SVO's benefits from that. I finally got the inspiration to try it out, and went for a SVO that is commonly used: the percentage:

The benchmark serializes an array of 1000 decimals / 1000 percentages, where most of them will have 4 digits, and some 5. With the new implementation we come close to writing raw decimals. Obviously we are slower, as we add 3 characters compared to a raw number in JSON. The durations are per SVO.

| Method     | Categories           | Mean     | Ratio |
|----------- |--------------------- |---------:|------:|
| Decimal    | JSON Deserialization | 42.65 ns |  1.00 |
| Percentage | JSON Deserialization | 52.71 ns |  1.24 |
|            |                      |          |       |
| Decimal    | JSON Serialization   | 25.65 ns |  1.00 |
| Percentage | JSON Serialization   | 39.77 ns |  1.55 |

The old implementations were 6 times (deserialization) and 13 times (serialization) slower.

The only downside for introducing this now is that I had to un-`seal`  the `Write` and `Read` methods of the `SvoJsonConverter<TSvo>` base class.

### DIFF
--- a/src/Qowaiv/Json/PercentageJsonConverter.cs
+++ b/src/Qowaiv/Json/PercentageJsonConverter.cs
@@ -57,7 +57,7 @@ public class PercentageJsonConverter : SvoJsonConverter<Percentage>
     ///  2 bytes for quotes
     ///  1 byte  for percentage sign
     ///  1 byte  for decimal seperator
-    ///  1 byte  for minius sign.
+    ///  1 byte  for minus sign.
     /// </remarks>
     public override void Write(Utf8JsonWriter writer, Percentage value, JsonSerializerOptions options)
     {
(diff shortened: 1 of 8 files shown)


src/Qowaiv/Json/PercentageJsonConverter.cs
```diff
@@ -21,6 +27,52 @@ public class PercentageJsonConverter : SvoJsonConverter<Percentage>
...
+        Span<byte> buffer = stackalloc byte[34];
+        buffer[0] = Quote;
+        Utf8Formatter.TryFormat(dec, buffer[1..], out var length);
+        buffer[++length] = Sign;
+        buffer[++length] = Quote;
+        writer.WriteRawValue(buffer[..++length], skipInputValidation: true);
+    }
+
+    private const byte Quote = (byte)'"';
+    private const byte Sign = (byte)'%';
```
Sjaaky: Sign had me confused. I thought is was about the Minus.
```suggestion
    private const byte Percent = (byte)'%';
```
or
```suggestion
    private const byte PercentageSign = (byte)'%';
```

Sjaaky (approved):
lgtm

PR #587, opened 2026-07-21
Amount JSON serialization speed improvements

Corniel:
The improvements are not as impressive as with the percentages #583 , but 20% is still okay.

| Method  | Categories      | Mean     | Ratio |
|-------- |---------------- |---------:|------:|
| New | Deserialization | 49.01 us |  1.00 |
| Old | Deserialization | 57.06 us |  1.17 |
|         |                 |          |       |
| New | Serialization   | 27.10 us |  1.00 |
| Old| Serialization   | 32.22 us |  1.19 |

### DIFF
--- a/src/Qowaiv/Json/Financial/AmountJsonConverter.cs
+++ b/src/Qowaiv/Json/Financial/AmountJsonConverter.cs
@@ -1,6 +1,8 @@
 #if NET8_0_OR_GREATER
 
 using Qowaiv.Financial;
+using System.Buffers.Text;
+using System.Text.Json;
 
 namespace Qowaiv.Json.Financial;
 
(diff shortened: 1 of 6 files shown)


src/Qowaiv/Json/Financial/AmountJsonConverter.cs
```diff
@@ -1,13 +1,26 @@
...
 
 using Qowaiv.Financial;
+using System.Buffers.Text;
+using System.Text.Json;
 
 namespace Qowaiv.Json.Financial;
 
 /// <summary>Provides a JSON conversion for an amount.</summary>
 [Inheritable]
 public class AmountJsonConverter : SvoJsonConverter<Amount>
```
JurJean, line 11: Does it still make sense to use this base class after overriding Read and Write?

JurJean (approved):

src/Qowaiv/Json/Financial/AmountJsonConverter.cs
Corniel: Yes it does. First of all, otherwise it would be a breaking change to drop it (I do not want a breaking change for things like that, at least not unless I have other breaking changes), and secondly, there still is a fallback path that uses it. (deserializing from string)

PR #603, opened 2026-08-26
Let Dependabot pull requests repair their own lock files

Gemberkoekje:
Dependabot leaves `packages.lock.json` broken on every NuGet update. So far that has been fixed by hand — #557, #578, #580, #584 and #598 all carry a second `lock files` commit; #576 didn't get one and was merged red.

Two upstream bugs, both in dependabot-core:

- [`LockFileUpdater`](https://github.com/dependabot/dependabot-core/blob/main/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Updater/LockFileUpdater.cs) has no callers, so nothing regenerates the lock files.
- `SdkProjectDiscovery` restores projects with more than two target frameworks once *per* framework, with `/p:TargetFramework=` pinned. Those run in parallel and race on the same lock file, and the winner leaves only its own framework behind. Hence NU1004 on exactly our seven multi-target projects, while the six single-target ones come out fine.

The NU1004 message reads back to front, which doesn't help: the list it labels *"Lock file target frameworks"* is what the project targets, and *"Project target frameworks"* is the single one left in the lock file.

This workflow runs `dotnet restore qowaiv.slnx --force-evaluate` on Dependabot PRs and pushes the result back. Verified on #600 and #601: after regeneration the diff against master is only the version bump, and a locked-mode Release build passes with 0 errors.

One wrinkle: a push made with `GITHUB_TOKEN` doesn't start a new workflow run, so Build won't re-run on the corrected commit — the PR just ends up with no Build result. Master has no required status checks, so that doesn't block the merge, and the push to master afterwards builds it. If we ever want the check back, swap the token for a PAT stored as a **Dependabot** secret (Actions secrets aren't visible to Dependabot runs).

To unblock #600 and #601: merge this, then `@dependabot rebase` on each.

Alternative, if you'd rather not run a workflow: get every project to two target frameworks or fewer. That's the threshold (`MaximumParallelTargetFrameworkRestores = 2`) at which Dependabot switches to a single unpinned restore and gets the lock files right on its own.

### DIFF
--- a/.github/workflows/dependabot-lock-files.yaml
+++ b/.github/workflows/dependabot-lock-files.yaml
@@ -11,9 +11,6 @@ on:
       - "**/*.csproj"
       - "**/packages.lock.json"
 
-permissions:
-  contents: read
-
 concurrency:
   group: dependabot-lock-files-${{ github.event.pull_request.number }}
   cancel-in-progress: true
@@ -23,22 +20,12 @@ jobs:
     name: Regenerate lock files
     if: github.actor == 'dependabot[bot]' && github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
-      # Has to be a Dependabot secret; GITHUB_TOKEN is read-only in Dependabot runs
-      # and pushes made with it do not trigger Build.
-      - name: Verify token
-        env:
-          LOCK_FILES_TOKEN: ${{ secrets.LOCK_FILES_TOKEN }}
-        run: |
-          if [ -z "$LOCK_FILES_TOKEN" ]; then
-            echo "::error::Missing Dependabot secret LOCK_FILES_TOKEN (fine-grained PAT, Contents: read and write)."
-            exit 1
-          fi
-
       - uses: actions/checkout@v7
         with:
           ref: ${{ github.event.pull_request.head.ref }}
-          token: ${{ secrets.LOCK_FILES_TOKEN }}
 
       - name: Setup .NET core versions
         uses: actions/setup-dotnet@v6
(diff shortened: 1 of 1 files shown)


Corniel:
That's quite a long change. Can you try to simply it? And what is so different about https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers that there it works with hardly any custom setup?

.github/workflows/dependabot-lock-files.yaml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      # Has to be a Dependabot secret; GITHUB_TOKEN is read-only in Dependabot runs
+      # and pushes made with it do not trigger Build.
+      - name: Verify token
+        env:
+          LOCK_FILES_TOKEN: ${{ secrets.LOCK_FILES_TOKEN }}
+        run: |
+          if [ -z "$LOCK_FILES_TOKEN" ]; then
+            echo "::error::Missing Dependabot secret LOCK_FILES_TOKEN (fine-grained PAT, Contents: read and write)."
```
Corniel: PAT?

Corniel:
This looks like a huge improvement, compared to the first commit/attempt. Easy(er) to follow. Lets try it out.

Corniel (approved):

.github/workflows/dependabot-lock-files.yaml
Gemberkoekje: De action heeft geen rechten om een branch te updaten by default, het is read only. Dus hebben we een PAT nodig die de action wat extra rechten geeft.

Gemberkoekje: Ik dubbelcheck nog even of dit echt nodig is.

Gemberkoekje: Ik wil het eerst zonder proberen, als het niet werkt kunnen we het altijd nog toevoegen.

PR #607, opened 2026-08-27
Fix the Dependabot lock file workflow trigger

Gemberkoekje:
Two problems from the first runs on #604.

**It skipped when you force-pushed.** `github.actor` is the actor of the *event*, not the author of the PR, so a human rebasing or reopening a Dependabot PR made the gate false and the job skipped. Now gated on `github.event.pull_request.user.login`.

**It failed when Dependabot abandoned the PR.** Sequence on #604: Dependabot force-pushed at 10:21:26, commented "no longer updatable" at 10:21:30, closed at 10:21:31 and deleted the branch at 10:21:33 — the checkout's fetch started right after and the ref was gone. Added a `git ls-remote` check so the job no-ops instead of failing.

Nothing else changed. #605 and #606 show the problem is still live: `+0 -204` on `src/Qowaiv.TestTools/packages.lock.json`, so Dependabot is now emptying lock files rather than just collapsing them.

### DIFF
--- a/.github/workflows/dependabot-lock-files.yaml
+++ b/.github/workflows/dependabot-lock-files.yaml
@@ -18,16 +18,29 @@ concurrency:
 jobs:
   lock-files:
     name: Regenerate lock files
-    if: github.actor == 'dependabot[bot]' && github.event.pull_request.head.repo.full_name == github.repository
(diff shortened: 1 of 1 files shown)


Corniel (approved):
Let's try it

PR #621, opened 2026-09-10
Add Percentage.Complement()

Corniel:
When reviewing the code of a co-worker, I saw this extension method (there called `Inverse()`). After some thoughts (advise of Gemini) I decided to go for `Complement` to get the `100.Percent() - value` of a percentage.

### DIFF
--- a/specs/Qowaiv.Specs/Percentage_specs.cs
+++ b/specs/Qowaiv.Specs/Percentage_specs.cs
@@ -930,7 +930,7 @@ public class Can_get_100_percent_based_on_percentage
     }
 }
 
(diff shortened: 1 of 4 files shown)


specs/Qowaiv.Specs/Percentage_specs.cs
```diff
@@ -930,6 +930,22 @@ public void _ushort()
     }
 }
 
+public class can_be_tranformed
```
Sjaaky:
```suggestion
public class can_be_transformed
```

Sjaaky (approved):
lgtm

